PR #3, opened 2022-07-13
Fix logs prefix

m4ra:
Instead of setting a universal `logging.basicConfig`, we add a formatter to the parent logger of the judoscale app. We can get the parent logger by giving the name of the app with `logger = logging.getLogger('judoscale')`.

The general logs of a django application can use either the default logging of Django, or the developers' logging added in the `settings.py`.
 I added custom logging in the django_sample/version 4.0 for compact logging of the django application, where the Django requests have now their own format:
 
django_4

While I left the default Django logging for the django_2_2_21, which logs output are:
django_2_2_21

The logs coming from Django when we hit localhost:8000 -> `[13/Jul/2022 17:43:14] "GET / HTTP/1.1" 200 10697`

and from localhost:5100 with our defined compact logging from settings.py-> `INFO "GET / HTTP/1.1" 200 10697`


adamlogic:
Thanks @m4ra! I'll take a look tomorrow.

src/judoscale/core/config.py
```diff
@@ -25,12 +23,17 @@ def for_report(self):
         }
 
     def _prepare_logging(self):
+        logger = logging.getLogger('judoscale')
+        print(self.log_level)
```
adamlogic, line 27: I assume this line was left in from debugging?

src/judoscale/core/config.py
```diff
@@ -25,12 +23,17 @@ def for_report(self):
         }
 
     def _prepare_logging(self):
+        logger = logging.getLogger('judoscale')
```
adamlogic, line 26: Can you help me understand the difference between this and `logging.getLogger(__name__)`? What is `__name__`? Is "judoscale" special in some way, or is this just an arbitrary name?

src/judoscale/core/config.py
```diff
@@ -25,12 +23,17 @@ def for_report(self):
...
+        print(self.log_level)
         log_level = logging.getLevelName(self.log_level.upper())
-        logging.basicConfig(
-            level=log_level,
-            format="%(levelname)s - [Judoscale] %(message)s",
-            force=True,
-        )
+        logger.setLevel(log_level)
```
adamlogic, line 29: Will this impact the log level of logs emitted outside of our package (such as logging from the Django app), or is it isolated to our package?

src/judoscale/core/config.py
```diff
@@ -25,12 +23,17 @@ def for_report(self):
...
+
+        stdout_handler = logging.StreamHandler()
+        stdout_handler.setLevel(logging.DEBUG)
```
adamlogic, line 32: I think this should be using the configured log level (`self.log_level`).

sample-apps/django_sample/django_sample/settings.py
```diff
@@ -130,3 +130,33 @@
...
+            'formatter': 'compact'
+        },
+    },
+    'loggers': {
+        'django': {
+            'handlers': ['console'],
+            'propagate': True,
+        },
+    }
+}
```
adamlogic, line 162: Does this just configure the "compact" logging? If so, I think it's probably unnecessary for this PR.

In the issue I'd asked for "Add logging to the sample app so we can reproduce the issue", and I should clarify that. What I mean is I'd like to see the sample app emit a custom log message (something like "hello world") so I can verify that there is no "[Judoscale]" prefix.

Right now the sample app just renders Django's default page, so I think it would be best to add a simple view for the root path (`/`) that just renders `<h1>Judoscale Sample App</h1>`. This view could contain the "hello world" log message mentioned above. Does this make sense?

adamlogic (changes requested):
@m4ra I think this is looking good. Just a few clarifying questions here.

src/judoscale/core/config.py
m4ra: `__name__` is a built-in attribute of Python that translates to the name of the file from which is called from. That's why sometimes at the end of a python file we see the expression `if __name__ == __main__: ` which means that if we run the file directly and do not import it as a module in another file, the __name__ translates to __main__ and then if that's the case we may want to do something particular or just call the functions and classes defined within that file. When we import the module in another file, this `__name__` translated to the path of the module.
I gave `judosacale` inside the getLogger because this is the library name, aka the root directory where django and core dirs are subdirectories. It is not just a random name. If you add the variable `name` in the formatter, which corresponds to `__name__`, you will see the names appear as `judoscale.core.adapter_api_client`, judoscale.django.apps` etc (it was the first screenshot I sent in the upwork messager). 
judoscale_logging_with_name

About the logging, I read in the docs, that if you want to start a parent logger and then children loggers can attach to it , you need to strat the logger with the package/module name where it is called for the first time.

> Logging is performed by calling methods on instances of the [Logger](https://docs.python.org/3/library/logging.html#logging.Logger) class (hereafter called loggers). Each instance has a name, and they are conceptually arranged in a namespace hierarchy using dots (periods) as separators. For example, a logger named ‘scan’ is the parent of loggers ‘scan.text’, ‘scan.html’ and ‘scan.pdf’. Logger names can be anything you want, and indicate the area of an application in which a logged message originates.

https://docs.python.org/3/howto/logging.html#logging-advanced-tutorial

sample-apps/django_sample/django_sample/settings.py
m4ra: Yes I can do that.

src/judoscale/core/config.py
m4ra: It is isolated because the logger here looks at the judoscale namespace in line 26 above.

src/judoscale/core/config.py
adamlogic: Actually, is this line even necessary since we're setting the level for `logger`?

src/judoscale/core/config.py
adamlogic: This is very helpful. Thanks!

src/judoscale/core/config.py
m4ra: No it is not, I will remove it, along with the print statement

sample-apps/django_sample/django_sample/settings.py
m4ra: I'd like to add a test for this view too. Is that ok?

src/judoscale/django/apps.py
```diff
@@ -27,14 +27,12 @@ def install_middleware(self):
             logger.info("Unable to install middleware")
             return False
 
-        logger.info("Installing middleware")
-
```
m4ra, line 31: This should log only if the middleware is missing, so after the expression: `if judoscale_middleware not in settings.MIDDLEWARE`

m4ra:
@adamlogic I added also a logging test for judoscale, and reworked existing tests to use unittesting.
You can run the tests from the root dir with `python -m unittest tests/judoscale/*_test.py`

Also I reworked some redundant code in the if/else expressions here https://github.com/judoscale/judoscale-python/pull/3/files#diff-ef42ed8c320c04983e10f80a63dd172ac140c5c183a28512879696d57c1fd855 

The new view comes from the new app `blog` in the sample django.
 To run the tests in the sample django, cd to `judoscale-python/sample-apps/django_sample` and do:
`python manage.py test`
When you run the sample django and visit localhost you see:

Screenshot from 2022-07-16 16-14-18
 Also I fixed the typo in the README of the django sample is `localhost:5100` the port,  not 5000.

sample-apps/django_sample/README.md
```diff
@@ -29,7 +29,7 @@ Run `bin/dev` to run the app in development mode. This will...
 
 Open https://judoscale-django.requestcatcher.com in a browser. The sample app is configured to use this Request Catcher endpoint as a mock for the Judoscale Adapter API. This page will monitor all API requests sent from the adapter.
 
-Start the app via `bin/dev`, then open http://localhost:5000. Continue to reload this page to collect and report more request metrics. You will see Judoscale Adatper API requests logged in Request Catcher.
+Start the app via `bin/dev`, then open http://localhost:5100. Continue to reload this page to collect and report more request metrics. You will see Judoscale Adatper API requests logged in Request Catcher.
```
adamlogic, line 32: `bin/dev` runs two processes (as [defined in `Procfile`](https://github.com/judoscale/judoscale-python/blob/7f24d65da376f0e1cf4f25a6a224fe5a5c73cd58/sample-apps/django_sample/Procfile)):

- `proxy` runs on port 5000. It proxies requests to port 5100.
- `web` runs the Django app on 5100.

The proxy adds the `HTTP_X_REQUEST_START` header so we can test the [request middleware](HTTP_X_REQUEST_START).

I think for you, the proxy is not running because you don't have Node installed (or at least `npx` is not available).

At any rate, I know the setup is a bit confusing, but I think port 5000 is correct here.

src/judoscale/core/config.py
```diff
@@ -1,16 +1,13 @@
...
-logger = logging.getLogger(__name__)
-
 
 class Config:
     def __init__(self):
         self.log_level = os.environ.get("LOG_LEVEL", "INFO")
         self.dyno = os.environ.get("DYNO", None)
         self.report_interval_seconds = 10
         self.api_base_url = os.environ.get("JUDOSCALE_URL", None)
-        self._prepare_logging()
```
adamlogic, line 13: Let's leave this here so the logger is set up correctly when `Config` is initialized even if `merge` is not called. We need it in both places because a caller could "merge" a different `log_level` into the config.

adamlogic (changes requested):
@m4ra A couple small changes, but otherwise this looks great!

How do I run the tests?

sample-apps/django_sample/README.md
m4ra: I installed npm so now npx can run, but I get this warning `WARN exec The following package was not found and will be installed: judoscale-adapter-proxy-server`.
But I will change the port back to 5000.

src/judoscale/core/config.py
m4ra: I can put it back, but then we get each log in double. Why not then just leave it only in the constructor __init__?

m4ra:
@adamlogic I mentioned how to run tests in my previous comment but here it is again:
You can run the tests from the root dir judoscale-python/ with `python -m unittest tests/judoscale/*_test.py`
To run the tests in the sample django, cd to judoscale-python/sample-apps/django_sample and do:
`python manage.py test`

sample-apps/django_sample/README.md
adamlogic: I think that message is expected the first time you run it. It should still work, though. Did you try hitting port 5000 after running it like this?

src/judoscale/core/config.py
adamlogic: > Why not then just leave it only in the constructor init?

Because the caller might provide a custom `log_level` when they call `merge`.

> I can put it back, but then we get each log in double.

Oh, I didn't realize that. I think the previous implementation of `_prepare_logging` was idempotent, so we could call it multiple times without multiple side effects. Would it be difficult for the new implementation to gracefully handle multiple calls?

adamlogic:
> I mentioned how to run tests in my previous comment but here it is again

Thank you, I overlooked it in the previous comment. I updated `tasks.json` in https://github.com/judoscale/judoscale-python/pull/3/commits/c64431be44290bfdb9ff3b9221ca66e700a54cba to facilitate running the tests in VS Code.

sample-apps/django_sample/README.md
m4ra: Yes, works at 5000 now as well and I can see the request time, so I put it back.
image

src/judoscale/core/config.py
```diff
@@ -25,12 +23,17 @@ def for_report(self):
...
-            format="%(levelname)s - [Judoscale] %(message)s",
-            force=True,
-        )
+        logger.setLevel(log_level)
+
+        if not logger.handlers:
+            stdout_handler = logging.StreamHandler()
+            fmt = "%(levelname)s - [Judoscale] %(message)s"
+            stdout_handler.setFormatter(logging.Formatter(fmt))
+            logger.addHandler(stdout_handler)
```
m4ra, line 34: @adamlogic this check of the existing handler fixed the duplicate logs. Because the _prepare_logging method is called once the object is initialized at the bottom, and then from merge method, it was re-adding a handler and when we do that the logs are printed again.

> Note If you attach a handler to a logger and one or more of its ancestors, it may emit the same record multiple times. 
https://docs.python.org/2/library/logging.html#logging.Logger.propagate

src/judoscale/core/adapter_api_client.py
```diff
@@ -12,7 +12,7 @@ def post_report(self, report):
             logger.debug("Posting metrics to Judoscale adapter API")
             requests.post(config.api_base_url + "/v1/metrics", timeout=5, json=report)
         except requests.RequestException as e:
-            logger.warn("Adapter API request failed - {}".format(e))
+            logger.warning("Adapter API request failed - {}".format(e))
```
m4ra, line 15: @adamlogic `warn` is deprecate

src/judoscale/core/config.py
adamlogic: Nice!

adamlogic (approved):
@m4ra This looks great to me. 👍 

Please merge when you're ready.

Issue #7, opened 2022-07-20
More better logging


adamlogic:
- [ ] `AdapterApiClient` logging should be _info_ level, not _debug_
- [ ] `AdapterApiClient` logging should include the number of metrics being reported (like the [Ruby adapter](https://github.com/judoscale/judoscale-ruby/blob/3571ec7b9a25093ff3223932d8a7a5178a324f8c/judoscale-ruby/lib/judoscale/reporter.rb#L80))
- [ ] Include PID in logs when starting reporter (this will help debugging parent/child process bugs)

PR #9, opened 2022-07-22
Start reporter from middleware

m4ra:


sample-apps/django_sample/Procfile
```diff
@@ -2,4 +2,5 @@
 proxy: npx judoscale-adapter-proxy-server
 
 # Run with --noreload to avoid multiple server processes (and multiple Judoscale reporters)
-web: python manage.py runserver $PORT --noreload
+web: gunicorn django_sample.wsgi:application --preload
+#web: python manage.py runserver $PORT --noreload
```
m4ra, line 6: @adamlogic shall I guess here should default to django's runserver than the gunicorn?

src/judoscale/core/adapter_api_client.py
```diff
@@ -9,8 +9,10 @@
 class AdapterApiClient:
     def post_report(self, report):
         try:
-            logger.debug("Posting metrics to Judoscale adapter API")
-            requests.post(config.api_base_url + "/v1/metrics", timeout=5, json=report)
+            url_metrics = config.api_base_url + "/v1/metrics"
+            logger.debug(
+                f"Posting metrics {report} to Judoscale adapter API {url_metrics}")
```
m4ra, line 14: Shall I keep this explicit debug info that shows what is actually is posted and where?

src/judoscale/core/reporter.py
```diff
@@ -12,14 +12,27 @@
...
+
+    @property
+    def is_running(self):
+        return self._running
+
     def _run_loop(self):
-        while True:
+        while self._running:
             self._report_metrics()
             time.sleep(config.report_interval_seconds)
```
m4ra, line 37: @adamlogic I made some changes here, a property method that we can check if a reporter is running, and a method to stop the thread but not in use anywhere yet.

sample-apps/django_sample/Procfile
adamlogic: I was thinking we would use gunicorn instead of runserver. Do you think there's benefit to testing with runserver?

src/judoscale/django/middleware.py
```diff
@@ -28,5 +29,16 @@ def __call__(self, request):
...
             logger.debug("queue_time={}ms".format(round(queue_time_ms, 2)))
+            self._get_reporter(reporter)
 
         return self.get_response(request)
+
+    def _get_reporter(self, reporter):
+        try:
+            if not reporter._thread.is_alive():
+                import os
+                pid = os.getpid()
```
adamlogic, line 40: Is `pid` being used anywhere?

src/judoscale/core/reporter.py
adamlogic: Is the intention to have the middleware use `is_running` instead of `reporter._thread.is_alive()`?

src/judoscale/core/adapter_api_client.py
m4ra: The output now is what this method returns https://github.com/judoscale/judoscale-python/blob/main/src/judoscale/core/reporter.py#L38
`DEBUG - [Judoscale] Posting metrics {'dyno': 'web.1', 'pid': 28438, 'config': {'log_level': 'DEBUG', 'report_interval_seconds': 2}, 'metrics': [[1658494001, 7.68, 'queue_time', None]]} to Judoscale adapter API https://judoscale-python.requestcatcher.com/v1/metrics`
It is indeed verbose. Where can I fetch this number of how many metrics? The `metrics` key has these data:     datetime -> as timestamp
    value -> the request time
    queue_name -> None (we don't get a queue_name, is that bcz we don't have celery jobs yet in place?)
    measurement -> "queue_time"
For debugging I think it would be helpful to leave also in the output which pid (aka process/worker) is reporting.

src/judoscale/core/reporter.py
m4ra: No, the intention is to help with testing. And in case we ever want to terminate a thread.

src/judoscale/django/middleware.py
m4ra: Yes in the [_build_reporter](https://github.com/judoscale/judoscale-python/blob/ed472cb72e2931cd237414988b70c56851114424/src/judoscale/core/reporter.py#L41) method, and it reflects the pid of the worker/process we see when the app starts

> 3:41:33 PM web.1   |  INFO - [Judoscale] **Starting reporter for process 27701**
> 3:41:33 PM web.1   |  DEBUG - [Judoscale] **Posting metrics {'dyno': 'web.1', 'pid': 27701**, 'config': {'log_level': 'DEBUG', 'report_interval_seconds': 2}, 'metrics': []} to Judoscale adapter API https://judoscale-python.requestcatcher.com/v1/metrics
> 3:41:33 PM web.1   |  [2022-07-22 12:41:33 +0000] [27701] [INFO] Starting gunicorn 20.1.0
> 3:41:33 PM web.1   |  [2022-07-22 12:41:33 +0000] [27701] [INFO] Listening at: http://0.0.0.0:5100 (27701)
> 3:41:33 PM web.1   |  [2022-07-22 12:41:33 +0000] [27701] [INFO] Using worker: sync
> 3:41:33 PM web.1   |  [2022-07-22 12:41:33 +0000] [27752] [INFO] **Booting worker with pid: 27752**
> 3:41:34 PM proxy.1 |  Listening on port 5000
> 3:41:34 PM proxy.1 |  Proxying requests to port 5100
> ...
> 3:41:51 PM web.1   |  DEBUG - [Judoscale] queue_time=7.29ms
> 3:41:51 PM web.1   |  INFO - [Judoscale] **Starting reporter for process 27752** (here it is the new reporter called from within the middleware when we hit the hello world page)
> 3:41:51 PM web.1   |  DEBUG - [Judoscale] **Posting metrics {'dyno': 'web.1', 'pid': 27752**, 'config': {'log_level': 'DEBUG', 'report_interval_seconds': 2}, 'metrics': [[1658493712, 7.29, 'queue_time', None]]} to Judoscale adapter API https://judoscale-python.requestcatcher.com/v1/metrics
> 3:41:51 PM web.1   |  Hello, world

adamlogic: I see. Is there any reason to call `os.getpid()` here, though? Looks like we're not using this `pid` local variable.

src/judoscale/core/adapter_api_client.py
adamlogic: > Where can I fetch this number of how many metrics?

`report["metrics"]` is a list, so I would just want the size of that list.

> we don't get a queue_name, is that bcz we don't have celery jobs yet in place?

Correct! If there's no queue name, we assume it to be _web queue time_, not _job queue time_.

> For debugging I think it would be helpful to leave also in the output which pid (aka process/worker) is reporting.

I agree. That was one of the logging improvements I had added to https://github.com/judoscale/judoscale-python/issues/7.

src/judoscale/django/middleware.py
m4ra: Oh right, my bad, I moved the pid in the log of the reporter.start() but forgot to delete it here.

sample-apps/django_sample/Procfile
m4ra: runserver is the native django's command to serve the application locally. Gunicorn is one option to serve django applications, [uwsgi](https://uwsgi-docs.readthedocs.io/en/latest/Configuration.html) is another popular option. Both are implemented usually for production. If the sample app is meant to be deployed in heroku by your clients to try out, then I'd leave the gunicorn.

sample-apps/django_sample/blog/tests.py
```diff
@@ -17,3 +19,13 @@ def test_app_logging(self):
...
+
+    def test_reporter_starts_for_each_process(self):
+        with self.assertLogs() as captured:
+            response = self.client.get('/')
+
+            self.assertEqual(response.status_code, 200)
+            # TODO patch the request header to return HTTP_X_REQUEST_START for test client
+            self.assertEqual(reporter.is_running, True)
+            print("METRICS", reporter.get_metrics())
+            self.assertEqual(len(reporter.get_metrics()), 0)
```
m4ra, line 31: I noticed that when running the tests, because there is no proxy server, but just the test client, we don't get metrics, I need to patch this for the test to give a length other than 0. But I can do it in a later PR. Let me know.

src/judoscale/core/reporter.py
```diff
@@ -38,10 +62,12 @@ def _build_report(self, metrics):
...
-            "pid": os.getpid(),
+            "pid": self.get_pid,
             "config": config.for_report(),
             "metrics": list(map(metric_to_list, metrics)),
         }
 
 
+import signal
 reporter = Reporter()
+signal.signal(signal.SIGILL, reporter.signal_handler)
```
m4ra, line 73: this signal will be used if we want to terminate the thread with a KILL process.

sample-apps/django_sample/Procfile
adamlogic: > If the sample app is meant to be deployed in heroku by your clients to try out

Not really. It's more for us (developers working on the adapter) to verify that the adapter is working as expected. So it's not really designed to be deployed to Heroku, but I do want it to _simulate_ how Django apps are commonly run on Heroku. So we'll leave the unicorn. :-)

sample-apps/django_sample/blog/tests.py
adamlogic: I'd prefer a later PR. 👍

src/judoscale/core/reporter.py
adamlogic: Good idea! Heroku [terminates dynos with SIGTERM](https://devcenter.heroku.com/articles/dynos#graceful-shutdown-with-sigterm), so I think we should handle that instead of (or in addition to) SIGKILL. wdyt?

Also, I assume `SIGILL` is a typo?

m4ra: uh yes, anyhow I removed it and replaced it with SIGTERM since you say this is what is used  by heroku. Locally threads are terminated with ctr+C (SIGINT) when we exit the local runserver which by default kills all daemon threads.

src/judoscale/django/middleware.py
```diff
@@ -28,5 +29,14 @@ def __call__(self, request):
 
             metrics_store.add(metric)
             logger.debug("queue_time={}ms".format(round(queue_time_ms, 2)))
+            self._get_reporter()
 
         return self.get_response(request)
+
+    def _get_reporter(self):
+        try:
+            if not reporter._thread.is_alive():
```
adamlogic, line 38: I'm confused how this works. If `reporter.start()` has not yet been called, I think `reporter._thread` will be `None`. I would expect calling `is_alive()` on `None` to raise an exception.

Do you see what I'm missing?

src/judoscale/django/middleware.py
```diff
@@ -28,5 +29,14 @@ def __call__(self, request):
 
             metrics_store.add(metric)
             logger.debug("queue_time={}ms".format(round(queue_time_ms, 2)))
+            self._get_reporter()
```
adamlogic, line 32: I think I'd like to move this implementation to a public `reporter.ensure_running()` method. We could call the same method from the [Django app initializer](https://github.com/judoscale/judoscale-python/blob/ce0b499333266f317dec0a059229c91a62d9d479/src/judoscale/django/apps.py#L23) rather than calling `reporter.start()` directly.

Does that make sense?

adamlogic (changes requested):
Looking good, @m4ra. Just a couple questions/requests.

src/judoscale/core/adapter_api_client.py
m4ra: That's a valid point. The thing is that the thread starts with the application preload, and then once gunicorn starts the worker this thread is stopped. I added some logging info on the thread itself to verify:

m4ra: But we should check if thread is alive inside the posting of the metrics so we don't get these logs of _posting 0 metrics..._

src/judoscale/django/middleware.py
m4ra: That's a valid point. The thing is that the thread starts with the application preload, and then once gunicorn starts the worker this thread is stopped. I added some logging info on the thread itself to verify:

> 9:53:47 AM web.1   |  INFO - [Judoscale] Installing middleware
> 9:53:47 AM web.1   |  INFO - [Judoscale] Starting reporter for process 16188
> 9:53:47 AM web.1   |  DEBUG - [Judoscale] Posting 0 metrics from 16188 to Judoscale adapter API https://judoscale-python.requestcatcher.com/v1/metrics
> 9:53:47 AM web.1   |  INFO - [Judoscale] **THREAD is <Thread(Thread-1, started daemon 140240101857024)> inside Reporter**
> 9:53:47 AM web.1   |  [2022-07-28 06:53:47 +0000] [16188] [INFO] Starting gunicorn 20.1.0
> 9:53:47 AM web.1   |  [2022-07-28 06:53:47 +0000] [16188] [INFO] Listening at: http://0.0.0.0:5100 (16188)
> 9:53:47 AM web.1   |  [2022-07-28 06:53:47 +0000] [16188] [INFO] Using worker: sync
> 9:53:47 AM web.1   |  [2022-07-28 06:53:47 +0000] [16238] [INFO] Booting worker with pid: 16238
> 9:53:48 AM proxy.1 |  Listening on port 5000
> 9:53:48 AM proxy.1 |  Proxying requests to port 5100
> 9:53:53 AM web.1   |  DEBUG - [Judoscale] Posting 0 metrics from 16188 to Judoscale adapter API https://judoscale-python.requestcatcher.com/v1/metrics
> 9:54:05 AM web.1   |  DEBUG - [Judoscale] queue_time=7.6ms
> 9:54:05 AM web.1   |  INFO - [Judoscale] **THREAD is <Thread(Thread-1, stopped daemon 140240101857024)> inside Middleware**
> 9:54:05 AM web.1   |  INFO - [Judoscale] Starting reporter for process 16238
> 9:54:05 AM web.1   |  DEBUG - [Judoscale] Posting 1 metrics from 16238 to Judoscale adapter API https://judoscale-python.requestcatcher.com/v1/metrics
> 9:54:05 AM web.1   |  INFO - [Judoscale] **THREAD is <Thread(Thread-2, started daemon 140240101857024)> inside Reporter**
> 9:54:05 AM web.1   |  Hello, world

m4ra: But we should check if thread is alive also before posting of the metrics so we don't get these logs of _posting 0 metrics..._

src/judoscale/django/middleware.py
m4ra: Yes, I will replace it. and in that case we need to check if thread is not None before checking its status.

src/judoscale/core/reporter.py
```diff
@@ -61,7 +74,8 @@ def _build_report(self, metrics):
 
         return {
             "dyno": config.dyno,
-            # TODO: Does each thread get its own PID?
+            # TODO: Does each thread get its own PID? Yes, except if a thread
+            # is stopped, its PID is recycled.
```
adamlogic, line 78: Thanks for answering this for me. We can delete this comment now. :-)

src/judoscale/core/reporter.py
```diff
@@ -49,7 +61,8 @@ def get_metrics(self):
 
     def _report_metrics(self):
         metrics = metrics_store.flush()
-        api_client.post_report(self._build_report(metrics))
+        if metrics:
```
adamlogic, line 64: We want to post a report even if `metrics` is an empty list.

This lets the Judoscale API know that the client (reporter) is active, just not receiving requests.

sample-apps/django_sample/Procfile
```diff
@@ -2,4 +2,4 @@
 proxy: npx judoscale-adapter-proxy-server
 
 # Run with --noreload to avoid multiple server processes (and multiple Judoscale reporters)
-web: python manage.py runserver $PORT --noreload
+web: gunicorn django_sample.wsgi:application --preload
```
adamlogic, line 5: I discovered an unexpected side effect of this change: The sample app uses the _published_ judoscale-python package instead of the _local_ code.

I tried making some logging changes locally, and I couldn't get those changes to show up in the sample app. When I reverted the Procfile change (to use `manage.py`), my logging changes showed up.

Do you know how we can pull in the local judoscale-python code rather than the published package? I need to do that so I can test these changes.

src/judoscale/core/reporter.py
```diff
@@ -12,17 +12,53 @@
...
     def _run_loop(self):
-        while True:
+        while self.is_running:
             self._report_metrics()
             time.sleep(config.report_interval_seconds)
 
+            if self._stopevent.is_set():
+                break
+
+        print(f"{self.get_pid} reports completed before exiting.")
```
adamlogic, line 57: Looks like some debugging still in here.

sample-apps/django_sample/Procfile
adamlogic: Are you seeing changes in this PR when you run `bin/dev`?

Here's what I see.

CleanShot 2022-08-01 at 11 05 25@2x

m4ra: Same with gunicorn, I added `TEST` in the logs:
image

adamlogic: I deactivated my venv (per your recommendation in Slack), and it's working now. Thanks!

adamlogic (approved):
All set here! 👍 👍

PR #11, opened 2022-08-04
Ci workflow unittests

m4ra:


adamlogic (approved):
Looks great! 🙌

m4ra:
@adamlogic because I changed the setuptools and the structure of the files, I think is best if you first try to build the package and do a test upload to pypi before merging to main?
Have you done a test package upload before? I could do that too if you don't have the time.
https://packaging.python.org/en/latest/guides/using-testpypi/

adamlogic:
> Have you done a test package upload before? I could do that too if you don't have the time.

I did once but it's been a while. It would be very helpful if you did. Thanks!

PR #13, opened 2022-09-16
Update API path

adamlogic:


adamlogic:
This will be replaced by https://github.com/judoscale/judoscale-python/issues/23

PR #14, opened 2022-09-21
Flask middleware

m4ra:


sample-apps/flask_sample/app.py
```diff
@@ -0,0 +1,28 @@
...
+import settings
+
+logger = logging.getLogger(__name__)
+
+
+app = Flask("DemoFlaskApp")
+config_obj = settings.BaseConfig
+app.config.from_object(config_obj)
+
+judoconfig.merge(getattr(config_obj, "JUDOSCALE", {}))
```
m4ra, line 15: @adamlogic ideally I'd like to move this line to judoscale/flask/settings.py, but flask has a number of ways to configure settings so it won't work with one way. Have to think about it a bit more
https://flask.palletsprojects.com/en/2.2.x/config/#configuration-best-practices

PR #15, opened 2022-09-22
Flask subpackage app

m4ra:


judoscale/flask/middleware.py
```diff
@@ -0,0 +1,52 @@
...
+        Remove decimal:
+            NGINX also reports this time as fractional seconds with millisecond
+            resolution, so removing the decimal gives us integer milliseconds
+            (same as Heroku).
+        Then, we calculate the timelapse between request start and finish,
+            log and send it to judoscale reporter.
+        """
+
+        request = Request(environ)
+        logger.debug(request)
```
adamlogic, line 33: Remove debug logging.

judoscale/flask/middleware.py
```diff
@@ -0,0 +1,52 @@
...
+            request_start_timestamp_ms = int(request_start_header)
+            current_timestamp_ms = now.timestamp() * 1000
+            queue_time_ms = current_timestamp_ms - request_start_timestamp_ms
+            metric = Metric(measurement="queue_time",
+                            datetime=now,
+                            value=queue_time_ms)
+
+            metrics_store.add(metric)
+            logger.debug("queue_time={}ms".format(round(queue_time_ms, 2)))
+            reporter.ensure_running()
```
adamlogic, line 50: Since this code is the same as the [Django middleware](https://github.com/rails-autoscale/judoscale-python/blob/ef49d78a9d4981bdd68c46f2b70948cb14dd93ef/judoscale/django/middleware.py#L24-L32), I'd like your thoughts on how we could extract it to `judoscale.core`.

The middleware should be responsible for getting the value of the header, but what we do with the header will be the same for each framework. I'm thinking of something like this:

```python
from judoscale.core.request_metrics import RequestMetrics

metric = RequestMetrics.get_queue_time_metric_from_header(request_start_header)
metrics_store.add(metric)
reporter.ensure_running()
```

`get_queue_time_metric_from_header` would handle removing non-digits from the string value, calculating the queue time, and returning the `Metric`. Maybe the logging as well.

We would move the comment block starting at line 21 into the core function as well. Thoughts?

adamlogic (changes requested):
This is looking good, and the demo app is working for me! 👍 

Could you also try to add a "Using Judoscale with Flask" section to the README? The writing doesn't need to be great—I will probably make changes—but I want to make sure the instructions are correct.

judoscale/flask/middleware.py
m4ra: Yes makes sense! I will move the block to core module. I think the logger of queue_time shall move within method `get_queue_time_metric_from_header` too.

m4ra: A new module RequestMetrics is now added in [judoscale/core/metric.py](https://github.com/rails-autoscale/judoscale-python/pull/15/files#diff-ce47cd840216c448cd5a9d25f3df2d203d07fe27600e83ec4119a013416089d4 ) and shared between the flask and django middleware.

README.md
```diff
@@ -56,6 +58,46 @@ JUDOSCALE = {
...
+
+```python
+# app.py
+from judoscale.flask.middleware import RequestQueueTimeMiddleware
+
+app = Flask("MyFlaskApp")
+app.wsgi_app = RequestQueueTimeMiddleware(app.wsgi_app)
+```
+
+This sets up the Judoscale middleware to capture request queue times.
```
adamlogic, line 73: ☝️ I added the bit above based on what I saw in the sample app. Does this correct? Anything you'd change?

adamlogic (approved):
This look great to me. I don't think I can test the sample app on Heroku until we publish the new package with Flask support, but it work well locally.

PR #19, opened 2022-10-28
Celery feature

m4ra:
WIP
**for locally testing** 
install system-wise rabbitmq and redis with:
```
sudo apt-get install rabbitmq-server
sudo apt-get install redis-server
```
and make sure those are running with 
```
systemctl status rabbitmq-server.service
sudo systemctl status redis-server.service
```
start these if not running
**Then**
navigate to `judoscale-python/sample-apps/celery_sample` and run:
`pip install -r requirements`
then in one terminal start celery with:
`celery -A celery_sample worker --loglevel=INFO -Q celery -E`
in another terminal run:
python monitor.py
and in a thrid terminal open a python shell and do:
```
(celery) joy% python -i                   
Python 3.9.0 (default, Jan 20 2022, 15:07:46) 
[GCC 6.3.0 20170516] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from celery_sample.celery import app
>>> from blog.tasks import add
>>> add.delay(1,3)
```
In the monitor.py terminal you will see timestamps of when the task was created, when was added to the queue, and the time in waiting after 5sec.. The task takes a sleep for 5'

**Next steps**
Think how to get the timestamp to the judoscale/core, perhaps the monitoring happens from there.
Any thoughts @adamlogic


sample-apps/celery_sample/requirements.txt
```diff
@@ -0,0 +1,10 @@
+asgiref
+certifi
+charset-normalizer
+Django
+gunicorn
+idna
+judoscale-python
+requests
+sqlparse
+urllib3
```
adamlogic, line 10: I had to also install `celery` and `redis` packages. Should they be in here?

adamlogic:
Thanks! I got it running once I added `celery` and `redis` to `requirements.txt`.



> Think how to get the timestamp to the judoscale/core, perhaps the monitoring happens from there.

Here's how I envision it (mirroring how the Ruby gem works):

- The celery package would "register" a "collector" with the reporter, so the reporter knows about it.
- When the reporter runs its loop, it would go through each "registered collector" to collect metrics. Today it only [calls `metrics.flush()`](https://github.com/rails-autoscale/judoscale-python/blob/a1fb9feb6f7f112e20c8d8e4c0d31e285a3a0cdd/judoscale/core/reporter.py#L63), but that only collects the web metrics. We also want it to "collect" metrics from the celery package.
- The celery collector would look at the oldest jobs in each queue to determine their queue time, and return those metrics.

sample-apps/celery_sample/requirements.txt
m4ra: Yes, I wil add these in the requirements

m4ra:
What does it mean register a collector? Celery has this method of monitoring for time creation of a task and the time it was received for execution. https://github.com/rails-autoscale/judoscale-python/pull/19/files#diff-eadb938603c480b87541ffcf4d70bc5ad3911db62dbd99966e33850fa1badaa7R59
My thought is to move this monitor file in judoscale/core and maybe inside the start_listening function, we call the _report_metrics. What do you think?
@adamlogic

adamlogic:
> What does it mean register a collector?

Here's the architecture of the Ruby gem:

Judoscale Adapter Architecture@2x-3

A "collector" is responsible for return metrics for a given framework. A `WebMetricsCollector` (Django/Flask) gets metrics by flushing the `MetricsStore` (which we currently do directly [in the reporter](https://github.com/rails-autoscale/judoscale-python/blob/a1fb9feb6f7f112e20c8d8e4c0d31e285a3a0cdd/judoscale/core/reporter.py#L63)), and a `JobMetricsCollector` gets metrics by inspecting the underlying data store for the queues.

The collector wouldn't need to have hard-coded knowledge about the specific collectors, just a mechanism for collectors to "register" themselves. This makes it easier to support many job frameworks (we want to support RQ in addition to Celery).

If this still isn't making sense, perhaps we should pause things for right now. I'm about to head out of town for the rest of the week, so we can get back on the same page early next week.

m4ra:
OK, I think we are in sync. I think of calling the reporter from the celery monitoring, which is in a way a job metrics collctor, since it calculates time spent in the queue. I will try it and we can discuss it again next week, on Tuesday? (next Monday am traveling back to Greece)

adamlogic:
Sounds good, @m4ra!

judoscale/celery/tasks_monitor.py
```diff
@@ -0,0 +1,87 @@
...
+            self.task_created[task.uuid] - current_timestamp_ms) * 1000
+        metric = Metric(
+            measurement="queue_time",
+            datetime=dt.fromtimestamp(self.task_created[task.uuid]),
+            value=queue_time_ms
+        )
+        logger.debug("queue_time={}ms".format(round(queue_time_ms, 2)))
+        if metric:
+            metrics_store.add_queue(metric)
+        reporter.ensure_running()
```
m4ra, line 52: @adamlogic metrics and reporter are added here.
The user needs to make a `monitor.py` file in the sample app and import the module `CeleryEventsHandler` from this file here (see the `monitor.py` below).

sample-apps/celery_sample/Procfile
```diff
@@ -0,0 +1,6 @@
+# The proxy server adds the X_REQUEST_START header for queue time calculations
+proxy: npx judoscale-adapter-proxy-server
+
+# Run with --noreload to avoid multiple server processes (and multiple Judoscale reporters)
+web: gunicorn celery_sample.wsgi:application --preload
+worker: celery -A celery_sample worker --loglevel=INFO -Q celery -E
```
m4ra, line 6: This line starts the celery workers

sample-apps/celery_sample/monitor.py
```diff
@@ -0,0 +1,7 @@
+from celery import current_app
+
+from judoscale.celery.tasks_monitor import CeleryEventsHandler
+
+
+events_handler = CeleryEventsHandler(current_app)
+events_handler.start_listening()
```
m4ra, line 7: So user need to run in addition python `monitor.py`
While metrics are gathered, and reporter is started, there are connections timed out to the `judoscale-python.requestcatcher.com`. See screenshot.
Screenshot from 2022-11-11 19-42-45
I wonder if the problem is with the ENV var `DYNO.` Because the ENV var `JUDOSCALE_URL` also was not picked up, and that is why I added it in the file judoscale/config,py:

sample-apps/celery_sample/README.md
```diff
@@ -0,0 +1,37 @@
+# Django Sample App
```
adamlogic, line 1: The README needs updated for how to setup and run the Celery sample app.

sample-apps/celery_sample/blog/tasks.py
```diff
@@ -0,0 +1,14 @@
+import time
+
+# from celery_sample.celery import app as celery_app
+# from celery_sample.celery import CeleryEventsHandler
+from celery import shared_task
+
+
+@shared_task(bind=True)
```
adamlogic, line 8: The sample app should have some way to trigger this task without running a python shell. For my testing, I added it in the blog index view so a task is enqueued each time I loaded localhost:5000/.

judoscale/celery/tasks_monitor.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class CeleryEventsHandler:
+    def __init__(self, celery_app, verbose_logging=False):
+        self._app = celery_app
+        self._state = celery_app.events.State()
+        self._logger = EventLogger()
+        self._verbose_logging = verbose_logging
+        self.task_started = {}
+        self.task_created = {}
```
adamlogic, line 33: It looks like we're continually adding entries to these objects and never clearing them out. Is that right?

judoscale/celery/tasks_monitor.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def announce_task_received(self, event):
+        self._state.event(event)
+        task = self._state.tasks.get(event["uuid"])
+        self.task_created[task.uuid] = task.timestamp
```
adamlogic, line 38: I'm really struggling to follow what's happening in this method. What is `task`, and how is it different from `event`? Where does `task.timestamp` come from? Why are we adding the timestamp to `task_created`, only to read from it again a few lines later?

judoscale/celery/tasks_monitor.py
```diff
@@ -0,0 +1,87 @@
...
+        # will keep track of this for us.
+        task = self._state.tasks.get(event["uuid"])
+
+        print("TASK FAILED: %s[%s] %s" % (
+            task.name, task.uuid, task.info(),))
+
+    def announce_task_succeeded(self, event):
+        self._state.event(event)
+        task = self._state.tasks.get(event["uuid"])
+        self._logger.log_task_status_change(task, event)
```
adamlogic, line 72: ☝️ Are these three methods doing anything?

sample-apps/celery_sample/monitor.py
adamlogic: Interesting... I did not get that error:


```
(.venv) celery_sample $ python monitor.py
[2022-11-11 21:23:25.083265] TASK-RECEIVED blog.tasks.add (STATE=RECEIVED, UUID=9049af8d-382b-4f14-a32b-c3bcac6eaf94)
INFO - [Judoscale] Starting reporter for process 14079
[2022-11-11 21:23:25.085320] TASK-STARTED blog.tasks.add (STATE=STARTED, UUID=9049af8d-382b-4f14-a32b-c3bcac6eaf94)
[2022-11-11 21:23:30.151912] TASK-SUCCEEDED blog.tasks.add (STATE=SUCCESS, UUID=9049af8d-382b-4f14-a32b-c3bcac6eaf94)
```

But I also don't see any log of anything being reported.

adamlogic (commented):
The main README needs instructions for how to install our Celery integration for Judoscale. There's a lot happening in the sample app, and I don't know how much of it is Judoscale-specific.

judoscale/django/apps.py
```diff
@@ -20,7 +20,7 @@ def ready(self):
             return
 
         self.install_middleware()
-        reporter.ensure_running()
+#        reporter.ensure_running()
```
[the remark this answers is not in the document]
m4ra: Testing to see if the timeouts were due to max requests being exceeded.

sample-apps/celery_sample/blog/tasks.py
m4ra: True, but resolving the reporting and debugging the task's error tracing is easier from the python shell. So that was my focus till now, to resolve the metrics reporting

sample-apps/celery_sample/README.md
m4ra: I will updated today.

judoscale/celery/tasks_monitor.py
m4ra: These methods are called in the `start_listening` method of the `CeleryEventHandler`, and they are reporting how the tasks are fairing. There are not needed eventually, there are part of the monitoring methods suggested by Celery, I temporarily left them in there for checking when tasks are created, received and completed to compare with the metrics gathered. When running `python monitor.py` they log the state of the tasks, using the `EventLogger` Class see the screenshot attached below, it outputs: 
TASK-RECEIVED blog.tasks.add (STATE=RECEIVED, UUID=...)
TASK-STARTED blog.tasks.add (STATE=STARTED, UUID=...)
TASK-SUCEEDED blog.tasks.add (STATE=SUCCESS, UUID=...)
...

judoscale/celery/tasks_monitor.py
m4ra: These methods are suggested by Celery's monitoring:
https://docs.celeryq.dev/en/stable/userguide/monitoring.html#events
Events are messages sent to the worker related to the tasks. We are adding the timestamps in the CeleryEventHandler class attribute so the other methods can read from these. I was using the time lag between task_created and task_started before for the metrics, and also in the method task_succeeded to check how long it took for the task to complete. I put it back now, so you understand better the reasoning.

m4ra: see https://github.com/rails-autoscale/judoscale-python/pull/19/files#diff-41f56399c272d79f41844d1600335d8c402315b04b982485aef065d8d967d6a0R72

PR #21, opened 2022-12-15
Migrate to Poetry

karls:
This PR migrates the project to use Poetry and removes `setuptools` and `pip` related code.

In addition to that, it introduces:
- [black](https://github.com/psf/black) for auto-formatting code
- [flake8](https://flake8.pycqa.org/en/latest/) for linting
- [isort](https://pycqa.github.io/isort/) for auto-sorting imports
- settings for VS Code tying all three of the above together, automatically


README.md
```diff
@@ -1,10 +1,14 @@
...
+This is the official Python adapter for
+[Judoscale](https://elements.heroku.com/addons/judoscale). You can use Judoscale
+without it, but this gives you request queue time metrics and job queue time
+(for supported job processors).
 
 ## Installation
 
-Add judoscale-python to your <code>requirements.txt</code> file:
+Add judoscale-python to your <code>requirements.txt</code> file or the
+equivalent:
```
adamlogic, line 11: It's kind of a bummer that the README must adhere to the 88-char line length, especially since Black isn't autoformatting it (at least for me). Did you have to make these changes manually, or does it automatically format for you? Note that Black *is* autoformatting Python files for me, just not the README.

Same question about comments within Python files. I noticed you broke them to adhere to the line length, but it doesn't appear to be automated.

I did a quick search to have Flake8 ignore comments, but I couldn't find anything. It's not something worth much time at all, but I wanted to bring it up in case you have an easy solution.

README.md
```diff
@@ -88,18 +96,53 @@ JUDOSCALE = {
...
+(1.2+) of Poetry installed.
+
+Clone the repo with
+
+```sh
+$ git clone git@github.com:judoscale/judoscale-python.git
+$ cd judoscale-python
+```
+
+Install dependencies with Poetry and activate the virtualenv
```
adamlogic, line 137: I think this is missing a step to install Poetry itself:

```sh
judoscale-python $ asdf current
crystal         1.5.0           /Users/adam/.tool-versions
nodejs          17.7.2          /Users/adam/.tool-versions
python          3.10.9          /Users/adam/src/judoscale-python/.tool-versions
ruby            3.0.3           /Users/adam/.tool-versions

judoscale-python $ python --version
Python 3.10.9

judoscale-python $ which poetry
/Users/adam/.asdf/shims/poetry

judoscale-python $ poetry install
No preset version installed for command poetry
Please install a version by running one of the following:

asdf install python 3.10.9

or add one of the following versions in your config file at /Users/adam/src/judoscale-python/.tool-versions
python 3.10.2
```

Looks like there's a [few ways to install Poetry](https://python-poetry.org/docs/), and I don't remember doing any of them when I ran your previous project that used Poetry. I feel like Poetry was already installed. 🤷‍♂️

adamlogic (commented):
This is looking great! Just a couple questions.

One other issue that I think is specific to my VS Code... when I launch my VS Code terminal, I'm still in a "venv". I don't see the same thing when I cd into this directly from Terminal.app—it's only VS Code.

CleanShot 2022-12-15 at 11 53 25@2x

Any idea what might be causing this, and how I can stop it?

README.md
karls: Oh.. the README doesn't have to have 88-char lines — that was me formatting it manually. I think I did it out of habit. Will revert this.

karls: Re: Flake8, Black and comments — yes, Black won't reformat comments and Flake8 will flag long lines. You could append `#  noqa: E501` which will cause Flake8 to ignore long lines. But I have always just reformatted long comments manually with `Vgq` (Vim keybindings).

adamlogic: I love that I'm still learning new Vim tricks after 15 years. 😂

No need to ignore long lines. It's a minor thing that I can live with.

README.md
karls: I think this is coming from `asdf`? https://github.com/asdf-vm/asdf/issues/914

FWIW I cannot replicate this behaviour. Also my `poetry` binary is in `/Users/karl/.local/bin/poetry`. I tried uninstalling Poetry completely and re-installing from Homebrew, which also worked without issues. The Homebrew version of Poetry was installed into `/opt/homebrew/bin/poetry`.

What version of Poetry do you have installed?

If I had to guess, I'd say your Poetry installation has somehow ended up somewhere where it shouldn't be. Perhaps you installed Poetry when you were trying out my previous project and the executable somehow ended up in the virtualenv for the that project?

AFAIK, Poetry should be installed "globally". Now, with `asdf`, "globally" is a slightly slippery term, but I don't think `poetry`-the-executable should end up in `~/.asdf/shims`.

I would try uninstalling Poetry completely, then re-installing either according to the [official instructions](https://python-poetry.org/docs/#installation) or from Homebrew.

adamlogic: > What version of Poetry do you have installed?

I couldn't tell because I couldn't run the executable. 😳 

But I think I'm good now. I got rid of my `poetry` shim in `~/.asdf/shims` by uninstalling Python 3.10.2. I still don't remember how I originally installed Poetry, but clearly I was using 3.10.2 at the time.

Anyway, now that the shim is gone, I installed Poetry using the [`curl` command on their site](https://python-poetry.org/docs/#installing-with-the-official-installer), and I'm good to go.

```
judoscale-python $ poetry --version
Poetry (version 1.3.1)
```

Thanks!

adamlogic:
> when I launch my VS Code terminal, I'm still in a "venv"

Looks like uninstalling Python 3.10.2 also fixed this! 🎉

adamlogic (approved):
This one's good to go once the Poetry install instructions are linked in the README. 

👍 Nice work!

karls:
> This is looking great! Just a couple questions.
> 
> One other issue that I think is specific to my VS Code... when I launch my VS Code terminal, I'm still in a "venv". I don't see the same thing when I cd into this directly from Terminal.app—it's only VS Code.
> 
> CleanShot 2022-12-15 at 11 53 25@2x
> 
> Any idea what might be causing this, and how I can stop it?

Oh I think the VS Code Python extension auto-activates the virtualenv when it detects that there is one present in the repo.

PR #22, opened 2022-12-15
Automate releases

karls:
This PR builds on #21 and introduces `.github/workflows/release.yml` which automates building the library, creating a release and publishing a new version on PyPI. This happens _only_ on `main` and _only_ when all tests have passed.

Note: this PR's branch was created from the branch for #21, which is why it includes commits from there. This should be ready to go once #21 is reviewed and merged.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,71 @@
+name: Release and publish
+
+on:
+  workflow_run:
+    workflows: [Run tests]
+    types: [completed]
+    branches: [main]
+
+jobs:
+  release-please:
+    name: Create a release
+    runs-on: ubuntu-22.04
+    outputs:
+      release_created: ${{ steps.release.outputs.release_created }}
+    steps:
+      - name: Release
+        id: release
+        uses: google-github-actions/release-please-action@v3
+        with:
+          release-type: python
+          package-name: judoscale-python
+          bump-minor-pre-major: true
+
+  publish:
+    name: Publish to PyPI
+    needs: release-please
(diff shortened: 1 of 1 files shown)


adamlogic:
@karls FYI I changed the base branch of this PR to `migrate-to-poetry` so this PR only includes the commits exclusive to this branch. GitHub will automatically change the base branch to `main` when `migrate-to-poetry` is merged.

.github/workflows/release.yml
```diff
@@ -0,0 +1,69 @@
...
+
+      - name: Install dependencies
+        run: poetry install --no-interaction --no-root
+        if: ${{ steps.cache-deps.outputs.cache-hit != 'true' }}
+
+      - name: Install library
+        run: poetry install --no-interaction
+
+      - name: Publish to PyPI
+        run: poetry publish --build --dry-run
```
adamlogic, line 69: Were you thinking we'd remove `--dry-run` before merging, after I have the secret in place?

karls: I think I'd like to see the workflow run successfully with `--dry-run` and with the secret in place before removing it. If you're OK with that, we could merge _with_ `--dry-run` and verify that everything works, then remove the flag when we're ready for a new release.

karls:
> GitHub will automatically change the base branch to `main` when `migrate-to-poetry` is merged.

I did not know that! Good stuff!

.github/workflows/release.yml
adamlogic: Sounds good! 👍

adamlogic (approved):

PR #26, opened 2022-12-27
Refactor Judoscale support for Flask into a Flask extension

karls:
This PR refactors what was previously custom middleware and setup for supporting Judoscale in a Flask app into a fully-fledged Flask extension. This is more familiar for Flask users, requires less setup and improves developer experience.

The explicit configuration step (via `judoconfig.merge`) is now replaced by an implicit configuration step in the extension's setup method `judoscale.init_app()`, which looks into the Flask application's config dictionary and uses the configuration found under `app.config["JUDOSCALE"]` to override its internal config variables.

```python
# Previously
from judoscale.flask.middleware import RequestQueueTimeMiddleware
from judoscale.core.config import config as judoconfig

app = Flask("MyFlaskApp")
app.wsgi_app = RequestQueueTimeMiddleware(app.wsgi_app)
judoconfig.merge(getattr(config_obj, "JUDOSCALE", {}))

# Now, with the global instance pattern
from judoscale.flask import Judoscale
app = Flask("MyFlaskApp")
judoscale = Judoscale(app)

# or with the application factory pattern
judoscale = Judoscale()
def create_app():
  app = Flask("MyFlaskApp")
  judoscale.init_app(app)
```

This approach is more in line with how Flask extensions are expected to work in general.

Additionally, the Flask sample app now uses the [application factory pattern](https://flask.palletsprojects.com/en/2.2.x/tutorial/factory/#the-application-factory) which is how most production Flask apps are set up (in my experience). The extension itself still supports both approaches — the app factory pattern and the global instance pattern.


judoscale/flask/judoscale.py
```diff
@@ -0,0 +1,26 @@
...
+    return None
+
+
+class Judoscale:
+    def __init__(self, app):
+        if app is not None:
+            self.init_app(app)
+
+    def init_app(self, app):
+        judoconfig.merge(app.config.get("JUDOSCALE", {}))
```
adamlogic, line 25: I think we might _also_ want to call `reporter.ensure_running()` here. We do [something similar](https://github.com/judoscale/judoscale-python/blob/bd0cdf3ef77137b539f1bc4497b97cc19bdc5e92/judoscale/django/apps.py#L23) in our Django integration to ensure the reporter is started as early as possible.

This is helpful for apps that get very little traffic, such as staging or demo apps. Customers often use these apps to experiment with Judoscale, and it's confusing for them if they install the adapter, but the adapter isn't recognized by the Judoscale service. This can happen if the app hasn't received traffic and the reporter hasn't started yet. By starting the reporter when the app boots, we ensure the service "sees" the adapter as early as possible.

Note that we'll still need to call `reporter.ensure_running()` when handling the request, because we need to handle forking web servers. I'm actually not sure if Flask can be run that way, but my understanding is that Gunicorn is a forking server similar to Puma in Ruby, and so a reporter started when the app boots will be running in a different process than the process that serves requests.

Does this all make sense?

adamlogic (commented):
This looks great, @karls! 

Would you also update the sample app with a `pyproject.toml` and update the README with Poetry instructions?

judoscale/flask/judoscale.py
karls: Makes perfect sense, yep. I've updated `.init_app()` to also ensure the reporter is running.

Gunicorn is a [forking web server](https://docs.gunicorn.org/en/stable/design.html#server-model), which — like you said — basically means that there will be a master process managing a handful of worker processes which are handling HTTP requests. For instance if there are 2 worker processes, we'd have 3 reporters running — one for the master process and one for each worker process. The master process will never execute the `report_metrics()` callback because it's not handling HTTP requests, so it'll continuously report empty metrics.

sample-apps/flask_sample/bin/test
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+poetry run python -m unittest
```
adamlogic, line 3: I think in the future we probably don't need tests for the sample apps. I know there were already tests in place here, so this is good for this PR.

.github/workflows/run-tests.yml
```diff
@@ -47,4 +47,4 @@ jobs:
         run: poetry install --no-interaction
 
       - name: Test with unittest
-        run: poetry run python -m unittest
+        run: poetry run python -m unittest discover -s tests
```
adamlogic, line 50: Does the README need a similar update?

The tests pass for me if I run `poetry run python -m unittest discover -s tests` as it's written here, but if I follow the README instructions, I get a failure:

```
judoscale-python $ poetry shell  
Spawning shell within /Users/adam/Library/Caches/pypoetry/virtualenvs/judoscale-python-gpoIQ17w-py3.10
judoscale-python $ emulate bash -c '. /Users/adam/Library/Caches/pypoetry/virtualenvs/judoscale-python-gpoIQ17w-py3.10/bin/activate'

(judoscale-python-py3.10) judoscale-python $ python -m unittest                                                                                               
E....
======================================================================
ERROR: judoscale.flask (unittest.loader._FailedTest)
----------------------------------------------------------------------
ImportError: Failed to import test module: judoscale.flask
Traceback (most recent call last):
  File "/Users/adam/.asdf/installs/python/3.10.9/lib/python3.10/unittest/loader.py", line 470, in _find_test_path
    package = self._get_module_from_name(name)
  File "/Users/adam/.asdf/installs/python/3.10.9/lib/python3.10/unittest/loader.py", line 377, in _get_module_from_name
    __import__(name)
  File "/Users/adam/src/judoscale-python/judoscale/flask/__init__.py", line 1, in <module>
    from judoscale.flask.judoscale import Judoscale
  File "/Users/adam/src/judoscale-python/judoscale/flask/judoscale.py", line 1, in <module>
    from flask import request
ModuleNotFoundError: No module named 'flask'


----------------------------------------------------------------------
Ran 5 tests in 0.013s

FAILED (errors=1)
```

adamlogic (approved):
I just have one question about the README and running tests, but otherwise I think this is good to go! I was able to follow your instructions on the sample app and verify the API calls. 👍

.github/workflows/run-tests.yml
karls: Ah yes, that's because `python -m unittest` is also picking up the sample app tests, but since the library and the sample apps have different virtualenvs, it cant' find "module named 'flask'".

The `-s tests` option scopes it to the `tests/` directory.

I think going forward we're going to use the "extras" feature in Python's packaging functionality, so we can actually convert this repo to use a single virtualenv which has all the deps for the sample apps as well.

judoscale/flask/judoscale.py
```diff
@@ -0,0 +1,27 @@
+from flask import request
+
+from judoscale.core.config import config as judoconfig
+from judoscale.core.metric import RequestMetrics
+from judoscale.core.metrics_store import metrics_store
+from judoscale.core.reporter import reporter
+
+
+def report_metrics():
```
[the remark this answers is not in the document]
karls: Agreed.

PR #29, opened 2023-01-16
Collector abstraction

karls:
This PR introduces the collector abstraction.

The collector abstraction relieves the reporter of the knowledge of _how_ or _where_ metrics are stored. The reporter only has to care about calling a particular method to get the metrics for any specific metrics collector.

The only collector that currently exists is for collecting web traffic metrics, as before. Except the reporter does not know anything about where web metrics are stored.

This work is lays the foundation for introducing more types of collectors, e.g job metrics collectors, etc.


judoscale/core/metric.py
```diff
@@ -1,45 +1,53 @@
...
 
 logger = logging.getLogger(__name__)
 
 
 @dataclass
 class Metric:
-    datetime: datetime
-    value: float
+    timestamp: float  # Unix timestamp in fractional seconds
+    value: float  # Metric value in milliseconds
```
adamlogic, line 13: This comment assumes a measurement of "queue_time". In the event of needing to support "queue_depth", `value` would actually represent the number of jobs instead of milliseconds.

```suggestion
    value: float
```

Afterthought: This really doesn't need to be a float. We don't care about sub-millisecond values, and if we need to measure "jobs" as in the scenario I just described, a float value doesn't make much sense.

No need to change this now, but just something to keep in mind.

judoscale/django/middleware.py
```diff
@@ -12,10 +12,9 @@ def __init__(self, get_response):
     def __call__(self, request):
         request_start_header = request.META.get("HTTP_X_REQUEST_START", "")
 
-        request_metric = RequestMetrics(request_start_header)
-        metric = request_metric.get_queue_time_metric_from_header()
-        if metric:
-            metrics_store.add(metric)
+        if metric := Metric.for_web(request_start_header):
+            if collector := reporter.find_collector(WebMetricsCollector):
```
adamlogic, line 16: It feels wrong to me that the middleware gets the collector from the reporter. I'm also concerned about the performance impact of running `find_collector` with every web request (which might be negligible).

Is there a way to pass the collector to the middleware's initializer, similar to how you [set it up for Flask](https://github.com/judoscale/judoscale-python/pull/29/files#diff-063a0f790822be3ab320856bb6c29bc1ebfbda52c7588a350d5005b63823d1e7R30)?

If not, is there an idiomatic way to have a singleton instance of `WebMetricsCollector`, since we'll only ever have one?

tests/test_metrics.py
```diff
@@ -0,0 +1,18 @@
...
+        # Heroku timestamp is in integer milliseconds
+        heroku_timestamp = str(round(time.time() * 1000))
+        metric = Metric.for_web(heroku_timestamp)
+        self.assertAlmostEqual(metric.value, 0, delta=1)
+
+    def test_nginx_metrics(self):
+        # Nginx timestamp is in float seconds, in the format "t=123.456"
+        nginx_timestamp = str(f"t={time.time()}")
+        metric = Metric.for_web(nginx_timestamp)
+        self.assertAlmostEqual(metric.value, 0, delta=1)
```
adamlogic, line 18: Nice!! 💯

adamlogic (changes requested):
Looking good! Just one concern about `find_collector`.

judoscale/core/metric.py
karls: That's a good spot and a fair point.

Updated it straight away as there's no need to kick the can down the road with this. 🙂

judoscale/django/middleware.py
karls: > It feels wrong to me that the middleware gets the collector from the reporter.
> Is there a way to pass the collector to the middleware's initializer, similar to how you set it up for Flask?

I'll have a deeper look and report back.

> I'm also concerned about the performance impact of running find_collector with every web request (which might be negligible).

I benchmarked this with Python's `timeit` module:
```
With a single collector, uncached:                                 92 nsec average
With four collectors, WebMetricsCollector first item, uncached:    94 nsec average
With four collectors, WebMetricsCollector last item, uncached:    460 nsec average
With 100 collectors, WebMetricsCollector first item, uncached:     92 nsec average
With 100 collectors, WebMetricsCollector last item, uncached:   12000 nsec average
Any variation from above, cached with @functools.cache:            55 nsec average
```

So one option is to cache the collector lookups. But I'm leaning towards seeing if there is better way, either via a singleton instance or some other mechanism.

karls: Turns out I had misunderstood how the Django middleware gets instantiated — it's done only once per process and each HTTP request is "handled" by the `__call__` method. This means we instantiate and store the collector in the middleware initialiser, as well as register it with the reporter, as in this commit: https://github.com/judoscale/judoscale-python/pull/29/commits/f46f6b5b0b62b0a064d7ab1c94924bd1de28cd13.

That looks like a better solution to me than the `.find_collector()` route.

tests/test_metrics.py
karls: I've tweaked the tests for Metric slightly — I added a tiny `sleep` into both tests (10ms each) to assert verify that latency calculation does take place correctly.

I've been running `watch -n 0.1 'python -m unittest discover -s tests'` in the background for hours and so far those parameters (10ms with 5ms delta) do not yield any false positives. But this is dependent on the load on the system, so if it ever becomes an issue in CI I think we'll just loosen the bounds a little bit.

karls: Tweaked the parameters slightly, because there were a couple of random failures with 10ms and 5ms delta.

judoscale/django/middleware.py
adamlogic: Yep, I think that's great! I also assumed the middleware was instantiated for each request, since that's how Rack middleware works in the Ruby world.

Is there a reason to keep `find_collector` around now?

adamlogic:
@karls Do you see this when you run `poetry install`?


```sh
judoscale-python $ poetry install
Installing dependencies from lock file
Warning: poetry.lock is not consistent with pyproject.toml. You may be getting improper dependencies. Run `poetry lock [--no-update]` to fix it.

No dependencies to install or update

Installing the current project: judoscale-python (0.1.1)

judoscale-python $ poetry lock [--no-update]
zsh: no matches found: [--no-update]

judoscale-python $ poetry lock --no-update 
Resolving dependencies... (0.5s)

The current project's Python requirement (>=3.8,<4.0) is not compatible with some of the required packages Python requirement:
  - flake8 requires Python >=3.8.1, so it will not be satisfied for Python >=3.8,<3.8.1

Because flake8 (6.0.0) requires Python >=3.8.1
 and no versions of flake8 match >6.0.0,<7.0.0, flake8 is forbidden.
So, because judoscale-python depends on flake8 (^6.0.0), version solving failed.

  • Check your dependencies Python requirement: The Python requirement can be specified via the `python` or `markers` properties
    
    For flake8, a possible solution would be to set the `python` property to ">=3.8.1,<4.0"

    https://python-poetry.org/docs/dependency-specification/#python-restricted-dependencies,
    https://python-poetry.org/docs/dependency-specification/#using-environment-markers
```

karls:
Whoops — I had a cached poetry.lock locally and didn't notice this issue. I do see it when I do `poetry install`.

The problem occurs because I relaxed the required Python version from `^3.10` to `^3.8` since this project can support lower Python versions. However it seems like `flake8` needs at least 3.8.1 and the resolver complains that the stated lower bound on the Python version is incompatible with `flake8 6.0.0`.

I've updated `pyproject.toml` so that at least 3.8.1 is required due to flake8, but in any production-like `pip install` scenario the lower bound on the Python version is still 3.8.0, because dev dependencies won't be included in the PyPI upload.

adamlogic (approved):
This is good to go! :shipit:

PR #31, opened 2023-01-26
Add Celery support

karls:
## This PR adds support for reporting Celery (Redis only) queue latency.

Changes to the existing code:

* Collectors now take `Config` as a constructor argument. This is essentially to make sure metrics are collected only on certain dynos.
* `Collector` is now a Protocol (an interface) and `MetricsCollector` is a plain Python class. The reporter expects to see a `Collector`, any concrete metrics collector will subclass `MetricsCollector`.
* More type annotations
* More comprehensive tests

New functionality:

* `CeleryMetricsCollector` class to iterate through Celery queues in Redis and collect queue latency (how long the oldest task has been waiting for).
* A user-callable function to instantiate and add the Celery collector to the reporter, which by virtue of importing also augments the Celery Task's properties with the `published_at` timestamp to be used downstream.

Observations/notes:
* Currently, the queues to report on are "discovered" using `SCAN`. Compared to `KEYS`, `SCAN` is more efficient from the Redis server perspective, but can be slower from the user's perspective. A more efficient way would be to use the [`active_queues()`](https://docs.celeryq.dev/en/stable/reference/celery.app.control.html#celery.app.control.Inspect.active_queues) method from `celery.app.control.Inspect`, but this only works if there are workers running, because workers register which queues they will be consuming from.
* The requestcatcher.com URL used in the sample apps fails fairly frequently with `WARNING - [Judoscale] Adapter API request failed - HTTPSConnectionPool(host='judoscale-celery.requestcatcher.com', port=443): Max retries exceeded with url: /v3/reports (Caused by ConnectTimeoutError(<urllib3.connection.HTTPSConnection object at 0x10328c8b0>, 'Connection to judoscale-celery.requestcatcher.com timed out. (connect timeout=5)'))`. Seems to be caused by a flaky service.


judoscale/core/metric.py
```diff
@@ -26,6 +26,18 @@ def as_tuple(self) -> Tuple[int, int, str, Optional[str]]:
...
+        """
+        Create a new Metric instance.
+
+        start_ms:
+            The start time the request or background job in milliseconds.
+        """
+        now = datetime.now(timezone.utc).timestamp()
+        latency_ms = max(0, int(now * 1000) - start_ms)
+        return cls(timestamp=now, value=latency_ms, **kwargs)
+
```
adamlogic, line 40: Nice refactor! 👍

judoscale/core/metrics_collectors.py
```diff
@@ -1,20 +1,55 @@
...
 from judoscale.core.metric import Metric
 from judoscale.core.metrics_store import MetricsStore
 
 
-class MetricsCollector(ABC):
-    @abstractmethod
+class Collector(Protocol):
     def collect(self) -> List[Metric]:
-        return []
+        ...
```
adamlogic, line 10: I take it `...` is another way to declare an abstract method?

judoscale/core/metrics_collectors.py
```diff
@@ -1,20 +1,55 @@
...
         return self.store.flush()
+
+
+class JobMetricsCollector(MetricsCollector):
+    def __init__(self, config: Config):
+        super().__init__(config=config)
+
+    @property
+    def should_collect(self):
+        return self.config.dyno_num == 1
```
adamlogic, line 55: Nice! I didn't remember talking about this, but it looks like you've handled all of https://github.com/judoscale/judoscale-python-issues/issues/1 in this PR.

sample-apps/celery_sample/settings.py
```diff
@@ -0,0 +1,19 @@
...
+    BASE_DIR = Path(__file__).parent.resolve()
+    # SECURITY WARNING: don't run with debug turned on in production!
+    DEBUG = True
+
+    SECRET_KEY = os.getenv("SECRET_KEY", "")
+
+    JUDOSCALE = {
+        "API_BASE_URL": "https://judoscale-celery.requestcatcher.com",
+        "LOG_LEVEL": "DEBUG",
+        "REPORT_INTERVAL_SECONDS": 2,
```
adamlogic, line 18:
```suggestion
        "REPORT_INTERVAL_SECONDS": 15,
```

I regret using a 2-second interval when I initially created the Django sample app. It makes so much noise when attempting to test! Let's start using longer intervals for sample apps so it's easier to see what's happening.

judoscale/celery/collector.py
```diff
@@ -0,0 +1,60 @@
...
+        logger.debug(f"Redis is at {self.redis.connection_pool.connection_kwargs}")
+        super().__init__(config=config)
+
+    @property
+    def queues(self) -> Set[str]:
+        """
+        Get all queues from Redis.
+        """
+        system_queues = {"unacked", "unacked_index"}
+        queues = set(self.redis.scan_iter(match="[^_]*", _type="list"))
```
adamlogic, line 33: It looks like this is run every time `collect` is run, is that right? I'm definitely concerned about hammering Redis if we have a bunch of Celery processes doing this every few seconds.

> A more efficient way would be to use the [active_queues()](https://docs.celeryq.dev/en/stable/reference/celery.app.control.html#celery.app.control.Inspect.active_queues) method from celery.app.control.Inspect, but this only works if there are workers running, because workers register which queues they will be consuming from.

Ah, and I suppose this would be a problem if worker dynos are scaled down to zero and we're relying on web dynos to report Celery queue time. Is that what you were thinking as well?

Another idea would be to only run this once when the collector is instantiated for an initial "snapshot" of the known queues. This would miss queues that are added after the process has been started, so maybe we also run it periodically, but less frequently than `collect` (maybe every 10th call to `collect` or something like that). Feels messy and I don't love it... just spitballing ideas.

Another idea is if `before_publish` had access to the collector, perhaps it could add the queue to the `collector.queues` set. I would think this would be a very fast operation since 99% of the time it would be a no-op (the queue is already present in the set).

adamlogic (changes requested):
This is great! I haven't tested it yet, but the code is looking really good. Readable even for a Rubyist. 😆 

Could you also add a README for the sample app with instructions on running locally and a section to the main repo README with instructions for installing the Celery adapter?

judoscale/core/metrics_collectors.py
karls: This is a shorthand for a built-in constant called [`Ellipsis`](https://docs.python.org/3/library/constants.html#Ellipsis) which I also learned about as I went along. It's pretty much like the `pass` keyword, except it can be used both in the function body as an expression and in the type signatures as well.

I initially went with `abc.ABC` as the mechanism for defining a collector interface. But the Python community seems to be leaning towards using `typing.Protocol` for interfaces as it leverages structural sub-typing rather than strict inheritance.

judoscale/core/metrics_collectors.py
karls: Yeah, it sort of just fell out so I thought why not do it straight away.

judoscale/celery/collector.py
karls: > Ah, and I suppose this would be a problem if worker dynos are scaled down to zero and we're relying on web dynos to report Celery queue time. Is that what you were thinking as well?

That's right. With Celery, the expectation seems to be that there's always at least one worker if you want to use the `Inspect` API.

> Another idea would be to only run this once when the collector is instantiated for an initial "snapshot" of the known queues.

The problem with this is that not all queues may be present when the snapshot is taken. Redis will remove the queue (key) if it's empty.

```sh
127.0.0.1:6379> keys "test_queue"
(empty array)
127.0.0.1:6379> lpush "test_queue" "test"
(integer) 1
127.0.0.1:6379> keys "test_queue"
1) "test_queue"
127.0.0.1:6379> lpop "test_queue" 1
1) "test"
127.0.0.1:6379> keys "test_queue"
(empty array)
127.0.0.1:6379>
```

If there isn't any work going on when the snapshot is taken, the set of queues reported back to the collector could be a subset (perhaps an empty set) of the actual queues the workers will see in production.

> Another idea is if before_publish had access to the collector, perhaps it could add the queue to the collector.queues set. I would think this would be a very fast operation since 99% of the time it would be a no-op (the queue is already present in the set).

I like this idea. The only rub with it is that because Celery allows enqueuing tasks with a routing key, rather than the queue name, we may not be able to determine which queue the task ends up on. I'll have a look tomorrow if there is a way know the queue name at the point of publishing.

> maybe we also run it periodically, but less frequently than collect (maybe every 10th call to collect or something like that)

This sounds like the most robust for getting an accurate set of queues, even though I'm with you on the messiness argument — doesn't feel super clean to me.

I'll do a bit of digging and come up with some approaches.

karls:
I think we may have to find an alternative to requestcatcher.com as well — getting a lot of timeouts. I've been trying out https://webhook.site/ which seems to work well. Open to alternatives though.

adamlogic:
Totally fine to switch from requestcatcher.com. It had plenty of downsides, I just hadn't really explored alternatives.

I like the simplicity of https://requestinspector.com/.

karls:
This approach — hooking into `before_task_publish` and storing queue names in the collector — seems to work quite well.

It works with the default setup with a single default queue, works with explicit routing via `.apply_async(queue=...)` and works with pre-configured routing via [`task_routes`](https://docs.celeryq.dev/en/latest/userguide/configuration.html#std-setting-task_routes) (even when using a function as the router).

It _does not_ work with task priorities, though. I've added a note in the README about this.

README.md
```diff
@@ -106,6 +106,38 @@ JUDOSCALE = {
 
 Note the [official recommendations for configuring Flask](https://flask.palletsprojects.com/en/2.2.x/config/#configuration-best-practices).
 
+# Using Judoscale with Celery and Redis
+
+> **NOTE 1:** The Judoscale Celery integration currently only works with the [Redis broker](https://docs.celeryq.dev/en/stable/getting-started/backends-and-brokers/index.html#redis).
+
+> **NOTE 2:** Using [task priorities](https://docs.celeryq.dev/en/latest/userguide/calling.html#advanced-options) is currently not supported.
```
adamlogic, line 113: I think we should clarify what "not supported" means here. My understanding is that users can still enqueue tasks with priorities, but Judoscale will be unaware of those priorities. Autoscaling is partitioned only on the queue names.

I think in an earlier conversation you said that if a task is enqueued without a queue name, it uses a default "celery" queue. Is that right?

If so, then if a user is _only_ using priorities for their tasks (no queue names), we'll still get queue time data, but it'll all be for a single "celery" queue. Sound right?

README.md
```diff
@@ -106,6 +106,38 @@ JUDOSCALE = {
...
+
+Judoscale can automatically scaling the number of Celery workers based on the queue latency (the age of the oldest pending task in the queue).
```
adamlogic, line 115:
```suggestion
Judoscale can automatically scale the number of Celery workers based on the queue latency (the age of the oldest pending task in the queue).
```

README.md
```diff
@@ -106,6 +106,38 @@ JUDOSCALE = {
...
+# Further setup
+# broker.conf.update(...)
+# ...
+
+judoscale_celery(broker)
+```
+
+This sets up Judoscale to periodically calculate and report queue latency for each Celery queue.
+
+Note that the queue names are captured via the [`before_task_publish` signal](https://docs.celeryq.dev/en/latest/userguide/signals.html#before-task-publish) and stored in the metrics collector.
```
adamlogic, line 133: Is this information relevant to the package consumer? It seems like internal-only details to me.

sample-apps/celery_sample/README.md
```diff
@@ -0,0 +1,32 @@
...
+## Run the app
+
+Run `bin/dev` to run the app in development mode. This will run `heroku local`, which starts:
+
+- a [tiny proxy server](https://github.com/judoscale/judoscale-adapter-proxy-server) that adds the `X-Request-Start` request header so we can test request queue time reporting; and
+- the sample app for testing Celery (with Flask).
+
+## How to use this sample app
+
+Open https://judoscale-celery.requestcatcher.com in a browser. The sample app is configured to use this Request Catcher endpoint as a mock for the Judoscale Adapter API. This page will monitor all API requests sent from the adapter.
```
adamlogic, line 30:
```suggestion
Open https://requestinspector.com/p/judoscale-celery in a browser. The sample app is configured to use this Request Catcher endpoint as a mock for the Judoscale Adapter API. This page will monitor all API requests sent from the adapter.
```

README.md
karls: D'oh!

README.md
karls: You're right. It's not relevant to the end user.

README.md
```diff
@@ -106,6 +106,38 @@ JUDOSCALE = {
...
+
+If you need to change any config options, you can pass a dictionary of Judoscale configuration options to override the default Judoscale config variables:
+
+```py
+judoscale_celery(broker, extra_config={"LOG_LEVEL": "DEBUG"})
+```
+
```
adamlogic, line 138: One other README change [near the top](https://github.com/judoscale/judoscale-python/pull/31/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R27):

```
- [x] Celery
```

😄

sample-apps/celery_sample/Procfile
```diff
@@ -0,0 +1,7 @@
+# The proxy server adds the X_REQUEST_START header for queue time calculations
+proxy: npx judoscale-adapter-proxy-server
+
+web: DYNO=web.1 poetry run gunicorn --preload 'app:create_app()'
+worker: DYNO=worker.1 poetry run celery -A app worker -c 1 --loglevel=INFO -Q low
+worker_high: DYNO=worker_high.1 poetry run celery -A app worker -c 1 --loglevel=INFO -Q high
+#web: flask run -p 5100
```
adamlogic, line 7:
```suggestion
```

sample-apps/celery_sample/Procfile
```diff
@@ -0,0 +1,7 @@
+# The proxy server adds the X_REQUEST_START header for queue time calculations
+proxy: npx judoscale-adapter-proxy-server
+
+web: DYNO=web.1 poetry run gunicorn --preload 'app:create_app()'
+worker: DYNO=worker.1 poetry run celery -A app worker -c 1 --loglevel=INFO -Q low
+worker_high: DYNO=worker_high.1 poetry run celery -A app worker -c 1 --loglevel=INFO -Q high
```
adamlogic, line 6: I was expecting to see the reporter running on all three processes, but it appears to only be running on `web.1`:

```
8:36:50 AM web.1         |  DEBUG - [Judoscale] Collecting metrics for queues ['high', 'low']
8:36:50 AM web.1         |  DEBUG - [Judoscale] queue_name=high, queue_time=10567ms
8:36:50 AM web.1         |  DEBUG - [Judoscale] queue_name=low, queue_time=10587ms
8:36:50 AM web.1         |  DEBUG - [Judoscale] Posting 10 metrics from 39677 to Judoscale adapter API https://requestinspector.com/inspect/judoscale-celery/v3/reports
8:36:53 AM web.1         |  DEBUG - [Judoscale] Collecting metrics for queues []
8:36:53 AM web.1         |  DEBUG - [Judoscale] Posting 0 metrics from 39487 to Judoscale adapter API https://requestinspector.com/inspect/judoscale-celery/v3/reports
8:37:06 AM web.1         |  DEBUG - [Judoscale] Collecting metrics for queues ['high', 'low']
8:37:06 AM web.1         |  DEBUG - [Judoscale] queue_name=high, queue_time=25681ms
8:37:06 AM web.1         |  DEBUG - [Judoscale] queue_name=low, queue_time=25808ms
8:37:06 AM web.1         |  DEBUG - [Judoscale] Posting 2 metrics from 39677 to Judoscale adapter API https://requestinspector.com/inspect/judoscale-celery/v3/reports
8:37:09 AM web.1         |  DEBUG - [Judoscale] Collecting metrics for queues []
8:37:09 AM web.1         |  DEBUG - [Judoscale] Posting 0 metrics from 39487 to Judoscale adapter API https://requestinspector.com/inspect/judoscale-celery/v3/reports
```

Will this still work if a user is running a "worker-only" app with no web process?

I'm also confused why only one of the two Gunicorn processes is reporting the Celery queues. Functionally that's fine since we only need one set of worker metrics, but can you help me understand why that's happening?

README.md
adamlogic: Thanks for the explanation. That is unfortunate. 😕

sample-apps/celery_sample/Procfile
karls: > Will this still work if a user is running a "worker-only" app with no web process?

Hmm.. no. It seems I didn't think of this use case. So web dynos can also scale down to 0? Can web and worker dynos be scaled down to 0 at the same time?

> I'm also confused why only one of the two Gunicorn processes is reporting the Celery queues. Functionally that's fine since we only need one set of worker metrics, but can you help me understand why that's happening?

Yep — Gunicorn starts a master process and N child processes to handle web requests. The master process itself will never see any web requests, it only supervises the child processes. The master process does not see the Celery queues because the collector gets the queue names from the `before_task_publish` handler, which is executed in the same process from which `.apply_async` is called — e.g the child process handling web requests.

Essentially, the reporter and the collectors exist in separate processes and are isolated from each other. The master process will never see metrics because it'll never handle any web requests and thus never enqueue any tasks.

karls: We can make the worker process also start a reporter, that's not a huge obstacle. The larger issue is how the reporter in the worker process will know which queues to report on, if we only know the queue name in the signal handler in the process we call `.apply_async()` in...

If the reporter has to run in the worker process as well then I think the current approach with getting queue names from the signal handler won't work.

karls: I think we can make this work by using the [`task-sent` event](https://docs.celeryq.dev/en/latest/userguide/monitoring.html#task-sent) and a slightly more complicated metrics collector, which needs to handle the `task-sent` event as well.

I'll prototype it to see if there are any obvious roadblocks.

karls: @adamlogic check out the latest changes. All three processes (web and two workers) are running the reporter now and reporting metrics as I think they should.

sample-apps/celery_sample/app/tasks.py
```diff
@@ -0,0 +1,17 @@
...
+import time
+
+from app.settings import BaseConfig
+from celery import Celery
+
+from judoscale.celery import judoscale_celery
+
+celery = Celery("CelerySampleApp", broker="redis://localhost:6379/0")
+
+judoscale_celery(celery, extra_config=BaseConfig.JUDOSCALE)
```
karls, line 11: I'm not a huge fan of this, if I'm being honest.

The end user should be able to override the Celery integration configuration, but I didn't want to make the mechanism specific on any particular web framework or anything like that. Even though in 99% of the cases it _will_ co-exist with a web framework I don't think we should be making assumptions about how the user will use this integration.

It'd be nice if the Celery integration could configure itself from the same configuration as the web app (`app.config` for Flask and `django.conf.settings` for Django), but I've left it like this for now.

adamlogic: I'm sort of surprised there's not already some kind of convention for this in Python.

In Ruby, for example, it's common for a library to have a global configuration mechanism that can be shared amongst multiple process. For `judoscale-ruby`, you [configure it](https://github.com/judoscale/judoscale-ruby#configuration) like this:

```ruby
Judoscale.configure do |config|
  config.report_interval_seconds = 5
end
```

As long as that code is executed by by the web and worker processes, they'll share the configuration. In a Rails app, that usually means you put it in an "initializer" (`/config/initializers/judoscale.rb`), and it will be auto-loaded by Rails.

But if there's no convention like that in Python, I think the way you have it here is fine.

judoscale/celery/collector.py
```diff
@@ -0,0 +1,79 @@
+import json
+from threading import Thread
+from typing import List, Optional, Set
+
+from celery import Celery
+from kombu import Connection
```
adamlogic, line 6: Do we need to include this in the project dependencies somewhere?

This actually reminds me that while this PR includes README instructions for *using* the Celery adapter, there's nothing specific to *installing* the Celery adapter. We'd discussed using the "extras" feature for each framework, right? For some reason I thought we'd done that for Django and Flask, but perhaps not.

judoscale/celery/__init__.py
```diff
@@ -0,0 +1,32 @@
...
+
+def judoscale_celery(celery: Celery, extra_config: dict = {}) -> None:
+    global before_publish_handler
+
+    celery.conf.task_send_sent_event = True
+
+    judoconfig.merge(extra_config)
+    collector = CeleryMetricsCollector(config=judoconfig, broker=celery)
+
+    before_publish_handler = before_task_publish.connect(before_publish(collector))
```
adamlogic, line 29: Can you help me understand the difference between "before publish" and "task sent", and why we're handling both?

karls: Whoops — I forgot to refactor this bit of code.

`before_task_publish` is the signal we hook into to add the `published_at` timestamp to the task's properties. This is signal only exists within the process.

`task-sent` is received by any process that sets up a `app.events.Receiver` to explicitly receive that event. It's more like pubsub than a signal sent and handled within the same process.

sample-apps/celery_sample/app/tasks.py
karls: Indeed — the Ruby approach is very nice. I'm not aware of any convention in Python that does this. One other option is to provide framework-specific ways to initialise this integration, e.g an extension for Flask, an app for Django, etc.

judoscale/celery/collector.py
karls: > Do we need to include this in the project dependencies somewhere?

If you're asking about `kombu`, then no, we don't need to include it in the project's dependencies. `kombu` is a dependency of `celery` and will be installed as a transitive dependency automatically.

> there's nothing specific to installing the Celery adapter

This library is currently still installable with the equivalent of `pip install judoscale-python`. Once we switch to the "extras" approach we'll update the README to include instructions for installing the specific integrations.

> We'd discussed using the "extras" feature for each framework, right? For some reason I thought we'd done that for Django and Flask, but perhaps not.

Yep, we have discussed it. But we haven't switched over to using extras yet. We could tackle this next, once the Celery work is merged.

sample-apps/celery_sample/app/tasks.py
karls: On second thought, it looks like the way Celery recommends initialising the broker for Django is [fairly simplistic](https://docs.celeryq.dev/en/stable/django/first-steps-with-django.html), so I'm not even sure adding a "Judoscale Django Celery" app would work. [Same for Flask](https://flask.palletsprojects.com/en/2.2.x/patterns/celery/).

judoscale/celery/collector.py
adamlogic: Thanks. I couldn't remember where we landed on that. I do think it would be good to tackle that next before we add another integration.

sample-apps/celery_sample/app/tasks.py
adamlogic: No problem. I'm okay with this config setup.

judoscale/celery/collector.py
```diff
@@ -0,0 +1,88 @@
...
+        except Exception as e:
+            logger.warning(f"Unable to get a task from queue: {queue}", exc_info=e)
+        return None
+
+    def collect(self) -> List[Metric]:
+        metrics = []
+        if not self.should_collect:
+            return metrics
+
+        logger.debug(f"Collecting metrics for queues {list(self.queues)}")
```
adamlogic, line 80: Found an issue here that I don't think I was clear about before. Even when nothing is pending in a given queue, we still want to report queue time for all known queues. In other words, reporting a queue time of 0 is better than not reporting the queue at all. This way Judoscale knows that the queue has been cleared, and it's not just waiting for the next metric.

In my testing, I've enqueued and processed jobs on the "low" and "high" queues. The queues are now empty. Here's what I see in the logs:

```
8:10:12 AM worker.1      |  DEBUG - [judoscale] Collecting metrics for queues ['low', 'high']
8:10:12 AM web.1         |  DEBUG - [judoscale] Collecting metrics for queues ['low', 'high']
8:10:12 AM worker_high.1 |  DEBUG - [judoscale] Collecting metrics for queues ['high', 'low']
8:10:12 AM worker.1      |  DEBUG - [judoscale] Posting 0 metrics from 6036 to Judoscale adapter API https://requestinspector.com/inspect/judoscale-celery-adamlogic/v3/reports
8:10:12 AM worker_high.1 |  DEBUG - [judoscale] Posting 0 metrics from 6037 to Judoscale adapter API https://requestinspector.com/inspect/judoscale-celery-adamlogic/v3/reports
8:10:12 AM web.1         |  DEBUG - [judoscale] Posting 0 metrics from 6035 to Judoscale adapter API https://requestinspector.com/inspect/judoscale-celery-adamlogic/v3/reports
```

Ideally, I still want each of these processes "posting 2 metrics" instead of "posting 0 metrics". They should be posting a 0 queue time for each of the known queues.

Does this make sense?

adamlogic (changes requested):

judoscale/celery/collector.py
karls: Yep, that makes sense. I've made the change to report latency on empty queues as well.

adamlogic (approved):
This works great, I think we're there!! Thanks for iterating so much on this. 🤘

PR #32, opened 2023-01-31
Update sample app configs

karls:
* Update reporting interval from two seconds to 15.
* Update request catcher service from [requestcatcher.com](https://requestcatcher.com/) to [requestinspector.com](https://requestinspector.com/).


adamlogic (commented):
👍 This works much better.

Can you also update the URL in the README for the sample apps?

PR #35, opened 2023-02-14
Installable as extras

karls:
Depends on #34.

This PR introduces a breaking change — the project name is now `judoscale` instead of `judoscale-python`.

The integrations will still be included if the user installs `judoscale` with `pip install judoscale`, but checking if the web framework and/or background task processing library is supported won't happen unless the user installs the suitable extras.


adamlogic (approved):
This looks great and seems to work great as well! 🙌  

Note that I updated `judoscale-python` to `judoscale` in a few more places in [f83dc73](https://github.com/judoscale/judoscale-python/pull/35/commits/f83dc734afa56bb8744bf8a1f09de6adc0e1af16), notably `release.yml`. Please give that a look before merging.

PR #46, opened 2023-03-17
fix: Handle minimum Redis version

karls:
`judoscale-python` relies on [`SCAN`](https://redis.io/commands/scan/) and its `TYPE` parameter to discover Celery queues. The `TYPE` parameters was only introduced in Redis 6.0.0.

This PR adds Redis version checking to prevent users using earlier Redis versions from seeing a confusing `ResponseError`. Instead, we raise a `RuntimeError` with an appropriate message.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ It is recommended to install the specific web framework and/or background job li
 
 ## Supported job processors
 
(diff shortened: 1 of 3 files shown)


adamlogic (approved):
👏👏👏

PR #48, opened 2023-03-28
feat: Add support for tracking busy jobs

karls:
This PR adds support for tracking busy jobs with RQ and Celery.

As has been the case with other features, Celery takes a bit more effort than RQ to make things work, and there are caveats. In this case, if `TRACK_BUSY_JOBS` is turned on, the Celery integration will report busy jobs counts only when workers are processing jobs. If there is no work, `self.inspect.active()` will return `None` and no busy metrics will be reported.

If this is a blocker and busy counts should be reported even if there is no work, I can take another look and find a workaround.


adamlogic (approved):
> If there is no work, self.inspect.active() will return None and no busy metrics will be reported.

I think it'll be okay. 👍

PR #54, opened 2023-04-04
feat: Add FastAPI support as ASGI middleware

karls:
This PR adds support for integrating Judoscale with FastAPI.

Since FastAPI uses Starlette under the hood, and Starlette is an ASGI framework/toolkit, this integration is packaged into ASGI middleware. The README has instructions for FastAPI, specifically, but since ASGI is a generic spec that [many web frameworks](https://asgi.readthedocs.io/en/latest/implementations.html) choose to implement, this integration mechanism will work for other web frameworks too.

I've added a simple FastAPI sample app. We may need to add examples of integrating with RQ or Celery, too.

Also made a change to `Config.update` to account for lowercase keys passed in via the `new_config` argument. This is for convenience, since some FastAPI users may be using the [`Pydantic.BaseSettings`](https://docs.pydantic.dev/usage/settings/#parsing-environment-variable-values) approach to configuration management.


adamlogic (approved):
Looks and works great!

> We may need to add examples of integrating with RQ or Celery, too.

Using our generic Celery instructions, here's what a user would see if they're using FastAPI and Celery. Do you think it's sufficient?

CleanShot 2023-04-04 at 17 42 51@2x

And similarly for RQ:

CleanShot 2023-04-04 at 17 45 41@2x

karls:
I think the instructions will work, yes. I've added a sample FastAPI Celery app as well, just in case — I wanted to make sure it all still works nicely together.

PR #56, opened 2023-04-17
chore: Add Python 3.12 to the build matrix

karls:
Currently blocked by https://github.com/celery/kombu/issues/1668


karls:
This was fixed in kombu 5.3.3 and celery 5.3.6.

PR #57, opened 2023-04-19
feat: Generic asgi request queue time middleware

karls:
This PR makes `RequestQueueTimeMiddleware` stop relying on Starlette's internals, which makes it more generic and re-usable for any app framework that adheres to the ASGI spec.

FastAPI users should now be installing Judoscale with `pip install 'judoscale[asgi]'`, instead of `pip install 'judoscale[fastapi]'`.

Middleware classes currently included for Quart, Starlette and FastAPI. Users can easily create their own if needed by following the instructions in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,27 +96,27 @@ Note the [official recommendations for configuring Flask](https://flask.palletsp
 Install Judoscale for FastAPI with:
 
 ```sh
-$ pip install 'judoscale[fastapi]'
+$ pip install 'judoscale[asgi]'
 ```
 
-Since FastAPI uses [Starlette](https://www.starlette.io/), an ASGI framework, the integration is packaged into [ASGI middleware](https://asgi.readthedocs.io/en/latest/specs/main.html#middleware). Import the middleare class and register it with your FastAPI app:
+Since FastAPI uses [Starlette](https://www.starlette.io/), an ASGI framework, the integration is packaged into [ASGI middleware](https://asgi.readthedocs.io/en/latest/specs/main.html#middleware). Import the middleware class and register it with your FastAPI app:
 
 ```py
 # app.py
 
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -125,9 +125,35 @@ This sets up the Judoscale extension to capture request queue times.
...
+
+```py
+# For Starlette, if you're using Starlette directly, without FastAPI
+from judoscale.asgi.middleware import StarletteRequestQueueTimeMiddleware
+
+# For Quart
+from judoscale.asgi.middleware import QuartRequestQueueTimeMiddleware
+```
+
+If your app uses a framework for which we have not provided a middleware class, but it implements the [ASGI spec](https://asgi.readthedocs.io/en/latest/specs/index.html), you can easily create your own version of the request queue time middleware.
```
adamlogic, line 143: 💯

pyproject.toml
```diff
@@ -36,7 +35,7 @@ pytest = "^7.2.2"
 [tool.poetry.extras]
 django = ["django"]
 flask = ["flask"]
-fastapi = ["fastapi"]
+asgi = []
```
adamlogic, line 38: With no dependencies, installing with the "asgi" extra doesn't actually do anything, right? Makes me wonder if it's worth having the extra, but I guess it keeps our installation process consistent (users are always using an extra depending on their framework). Is that what you're thinking?

adamlogic (approved):
Love it!

pyproject.toml
karls: That's exactly right. I was considering leaving this out, but it feels weird to install everything else as an extra, but not ASGI support...

I was going for consistency here. Functionally it doesn't do anything.

adamlogic: I'm on board with that!

PR #60, opened 2023-04-27
fix: Handle missing API_BASE_URL

### LINKED ISSUE #59: Default config for non heroku environments
The default config for non Heroku environments is currently set as follows:
https://github.com/judoscale/judoscale-python/blob/e101123c8fc29d61b71de67cfa1dcee194ee7e4b/judoscale/core/config.py#L53

so `API_BASE_URL` is set as an empty string.

The code that determines if Judoscale should run, though, seems to compare `API_BASE_URL` against `None`:

https://github.com/judoscale/judoscale-python/blob/e101123c8fc29d61b71de67cfa1dcee194ee7e4b/judoscale/django/apps.py#L20

We've been seeing Judoscale attempting to run locally on our dev machines, and pretty sure it's because this check is returning True.

Should the default config instead return `return cls(None, None, env)` so the `API_BASE_URL` is set to None instead of ""? Or alternatively should the check be `if judoconfig["API_BASE_URL"]:`? Happy to put up a pull request if this seems reasonable.

karls:
Fixes #59.

### DIFF
--- a/judoscale/asgi/middleware.py
+++ b/judoscale/asgi/middleware.py
@@ -3,6 +3,7 @@ from typing import Mapping
 
 from judoscale.core.adapter import Adapter, AdapterInfo
 from judoscale.core.config import config as judoconfig
+from judoscale.core.logger import logger
 from judoscale.core.metric import Metric
 from judoscale.core.metrics_collectors import WebMetricsCollector
 from judoscale.core.reporter import reporter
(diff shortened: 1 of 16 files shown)


karls:
@adamlogic there's some extra noise again due to updating sample apps as well. The first commit ([e4b9a77](https://github.com/judoscale/judoscale-python/pull/60/commits/e4b9a7702f4bbc8ef76397e90b8d6cc9c1ebe8fc)) is the library change, and the second commit ([bb764df](https://github.com/judoscale/judoscale-python/pull/60/commits/bb764df0eb56a44aef46c52ea525a6225c9d583f)) is the sample apps change. Just FYI. 🙂

adamlogic (approved):
👍 👍

PR #74, opened 2023-10-02
fix: judoscale.rq not getting registered in Django < 3.2

karls:
Django 3.2 introduced [automatic app discovery](https://docs.djangoproject.com/en/4.2/releases/3.2/#what-s-new-in-django-3-2). This PR fixes the `judoscale.rq` app not getting registered in Django, even if `judoscale.rq` was added to `INSTALLED_APPS`. The fix uses the same approach as for `judoscale.django`.

Also added a sample app for testing the exact setup as reported by a customer.


adamlogic:
Thanks @karls!

PR #75, opened 2023-10-08
fix: Redis client configuration for RQ

karls:
Heroku Data for Redis premium plans only offer a TLS connection to Redis, which uses self-signed certificates, which require you to configure the verify_mode SSL setting of your Redis client.

This PR fixes the inability to configure this in `JUDOSCALE["REDIS"]` configuration.


adamlogic:
@karls Thanks for the fix! Will we potentially have the same issue for Celery?

PR #85, opened 2024-06-03
fix: Improve automatic queue detection for Celery

karls:
In certain configurations the adapter library outputs a large volume of warning logs that look something like this:

> WARNING - [judoscale] Unable to find `published_at` in task properties for task ID None in queue e752fa70-f772-3c04-b05d-79b2a79ce766.reply.celery.pidbox.

This happens because `e752fa70-f772-3c04-b05d-79b2a79ce766.reply.celery.pidbox` is incorrectly detected as a user queue. Consequently, the adapter tries to inspect the queue to determine queue latency, but the items in there do not correspond to Celery tasks, which causes the warning log lines to get emitted by the adapter library.

This PR tightens up the mechanism by which the adapter tries to discover the Celery queues to monitor.

### DIFF
--- a/judoscale/celery/collector.py
+++ b/judoscale/celery/collector.py
@@ -73,12 +73,17 @@ class CeleryMetricsCollector(JobMetricsCollector):
         logger.debug(f"Redis is at {self.redis.connection_pool}")
 
         system_queues = {"unacked", "unacked_index"}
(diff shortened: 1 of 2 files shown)


adamlogic (approved):
👍 Thanks!

PR #97, opened 2024-12-12
Fix broken docs links

bennylope:
The URL with [`-ruby`](https://judoscale.com/docs/long-running-jobs-ruby#enable-long-running-job-support-in-the-dashboard) has gone the way of the Norwegian blue parrot.


adamlogic:
Thanks @bennylope!

PR #107, opened 2025-08-04
fix: Ensure reporter thread only starts once per process

carlosantoniodasilva:
### Testing

I tested by running the Django server (here with the Procfile changed to run 3 workers and 2 threads each) with `bin/heroku | grep Starting` to filter just the messages we cared about, and then executed `hey` against the server to hit it with a few concurrent requests: `hey 'http://localhost:5006/?sleep=t' -n 20 -c 5`

Behavior on main:

```
% bin/heroku | grep Starting
1:55:39 PM web.1   |  INFO - [judoscale] Starting reporter for process 11661
1:55:39 PM web.1   |  [2025-08-04 16:55:39 +0000] [11661] [INFO] Starting gunicorn 20.1.0
1:55:42 PM web.1   |  INFO - [judoscale] Starting reporter for process 11664
1:55:42 PM web.1   |  INFO - [judoscale] Starting reporter for process 11666
1:55:42 PM web.1   |  INFO - [judoscale] Starting reporter for process 11666
1:55:42 PM web.1   |  INFO - [judoscale] Starting reporter for process 11665
```

Behavior with fix:

```
% bin/heroku | grep Starting
2:03:53 PM web.1   |  INFO - [judoscale] Starting reporter for process 11906
2:03:53 PM web.1   |  [2025-08-04 17:03:53 +0000] [11906] [INFO] Starting gunicorn 20.1.0
2:03:54 PM web.1   |  INFO - [judoscale] Starting reporter for process 11909
2:03:54 PM web.1   |  INFO - [judoscale] Starting reporter for process 11911
2:03:54 PM web.1   |  INFO - [judoscale] Starting reporter for process 11910
```

### DIFF
--- a/judoscale/core/__init__.py
+++ b/judoscale/core/__init__.py
@@ -1,0 +1,21 @@
+import signal
+import sys
+
+from judoscale.core.reporter import reporter
+
+previous_handler = signal.getsignal(signal.SIGTERM)
+
+def graceful_shutdown(signum, frame):
+    reporter.stop()
+    # If there was a previous handler and it's not SIG_DFL or SIG_IGN, call it
+    if callable(previous_handler) and previous_handler not in (
+        signal.SIG_DFL,
+        signal.SIG_IGN,
+    ):
+        previous_handler(signum, frame)
+    else:
+        # If no previous handler, use default behavior (exit the process)
+        sys.exit(0)
+
+
+signal.signal(signal.SIGTERM, graceful_shutdown)
--- a/judoscale/core/reporter.py
+++ b/judoscale/core/reporter.py
@@ -1,6 +1,4 @@
 import os
-import signal
-import sys
 import threading
 import time
 from platform import python_version
@@ -141,21 +139,3 @@ class Reporter:
 
 
 reporter = Reporter(config=config)
-
-previous_handler = signal.getsignal(signal.SIGTERM)
-
-
-def graceful_shutdown(signum, frame):
-    reporter.stop()
-    # If there was a previous handler and it's not SIG_DFL or SIG_IGN, call it
-    if callable(previous_handler) and previous_handler not in (
-        signal.SIG_DFL,
-        signal.SIG_IGN,
-    ):
-        previous_handler(signum, frame)
-    else:
-        # If no previous handler, use default behavior (exit the process)
-        sys.exit(0)
-
-
-signal.signal(signal.SIGTERM, graceful_shutdown)
--- a/judoscale/core/reporter.py
+++ b/judoscale/core/reporter.py
@@ -23,6 +23,7 @@ class Reporter:
         self.config = config
         self._thread = None
         self._running = False
+        self._lock = threading.Lock()
         self._stopevent = threading.Event()
         self.collectors: List[Collector] = []
         self.adapters: List[Adapter] = []
@@ -65,8 +66,9 @@ class Reporter:
 
     def ensure_running(self):
         try:
-            if not self.is_running:
-                return self.start()
+            with self._lock:
+                if not self.is_running:
+                    return self.start()
         except Exception as e:
             logger.warning(f"{e.args} - No reporter has initiated")
             pass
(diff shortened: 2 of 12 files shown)


judoscale/core/__init__.py
```diff
@@ -0,0 +1,21 @@
...
+        signal.SIG_DFL,
+        signal.SIG_IGN,
+    ):
+        previous_handler(signum, frame)
+    else:
+        # If no previous handler, use default behavior (exit the process)
+        sys.exit(0)
+
+
+signal.signal(signal.SIGTERM, graceful_shutdown)
```
carlosantoniodasilva, line 21: I pulled this change (and a few other small test changes) from https://github.com/judoscale/judoscale-python/pull/106, only thing I did was remove the `signal_handler` there in favor of just `stop` since we don't car about the args.

judoscale/core/reporter.py
```diff
@@ -64,27 +63,23 @@ def start(self):
...
         self._thread.start()
-        self._running = True
 
     def ensure_running(self):
         try:
-            if not self.is_running:
-                return self.start()
+            with self._lock:
+                if not self.is_running:
+                    return self.start()
```
carlosantoniodasilva, line 71: Don't love adding a lock around this since it's hit on every request... but it's certainly safe.

I also tried with the pid check we do in Ruby, which I'm not sure it's 100% safe, but it actually seemed to work on all my tests... only problem is it doesn't check for the thread dying, but I guess that's also a problem on the Ruby version currently.

```
% bin/heroku| grep Starting
2:42:01 PM web.1   |  INFO - [judoscale] Starting reporter for process 17807
2:42:01 PM web.1   |  [2025-08-04 17:42:01 +0000] [17807] [INFO] Starting gunicorn 20.1.0
2:42:02 PM web.1   |  INFO - [judoscale] Starting reporter for process 17830
2:42:02 PM web.1   |  INFO - [judoscale] Starting reporter for process 17829
2:42:02 PM web.1   |  INFO - [judoscale] Starting reporter for process 17831
```

This is the basic diff for the logic:

```diff
diff --git a/judoscale/core/reporter.py b/judoscale/core/reporter.py
index dbb894b..b0bac59 100644
--- a/judoscale/core/reporter.py
+++ b/judoscale/core/reporter.py
@@ -21,6 +21,7 @@ class Reporter:

     def __init__(self, config: Config):
         self.config = config
+        self.pid = None
         self._thread = None
         self._running = False
         self._lock = threading.Lock()
@@ -66,9 +67,9 @@ class Reporter:

     def ensure_running(self):
         try:
-            with self._lock:
-                if not self.is_running:
-                    return self.start()
+            if (pid := os.getpid()) and self.pid != pid:
+                self.pid = pid
+                return self.start()
         except Exception as e:
             logger.warning(f"{e.args} - No reporter has initiated")
             pass
@@ -81,10 +82,6 @@ class Reporter:
     def is_running(self):
         return self._thread and self._thread.is_alive()

-    @property
-    def pid(self) -> int:
-        return os.getpid()
-
```

Perhaps that's good enough without adding a lock, since it's actually not a destructive problem if we end up running an extra reporter every now and then? (just not almost every time as it happens today with threading servers)

sample-apps/django_sample/blog/views.py
```diff
@@ -1,14 +1,22 @@
...
 def index(request):
     # Log message in level warning as this is Django's default logging level
     logger.warning("Hello, world")
+
+    if sleep_for := request.GET.get("sleep"):
+        try:
+            sleep_for = float(sleep_for)
+        except:
+            sleep_for = random.randint(0, 2)
+        time.sleep(sleep_for)
```
carlosantoniodasilva, line 18: Cherry picked these from the other branch too, and expanded them a bit to pass a number as param or use a random number.

carlosantoniodasilva (commented):
@adamlogic lmk what you think, especially around the lock vs the pid check in here.

judoscale/core/reporter.py
carlosantoniodasilva: I ran `hey` with different numbers of requests (1k, 10k) against main, this branch, and this branch applying the above diff, and there was no noticeable change.

Let's go with the lock version, and we can revisit if we hear any feedback otherwise.

PR #110, opened 2025-10-02
Handle Celery `eta`/`countdown` by skipping future dates and using past dates

carlosantoniodasilva:
If a task is enqueued with an eta, when looking for the oldest job we
must skip the task if the eta is still in the future, to make sure we
don't use a future-dated task with a potentially older published_at
timestamp to calculate our queue latency.

And if the task has an eta already in the past, we should use that
instead of our published timestamp, to calculate queue time.

This basically ignores the time between the task is published and when
it's actually meant to execute, considering the eta as the "enqueue"
time when available.

This should fix an issue where we sometimes see a spike of queue time
apparently related to the usage of `countdown`/`eta`, because that task
may end up back in the queue and we may pick it up as part of our
reporting, causing us to report a spike in queue time that goes away
within the next report or so. (that is because tasks with `eta` live in
the worker's memory most of the time.)

### DIFF
--- a/judoscale/celery/collector.py
+++ b/judoscale/celery/collector.py
@@ -1,8 +1,9 @@
 import json
 import time
+from datetime import datetime
 from collections import defaultdict
 from threading import Thread
-from typing import List, Optional, Set
+from typing import List, Optional, Set, Tuple
 
 from celery import Celery
 from kombu import Connection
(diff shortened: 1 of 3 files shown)


adamlogic (approved):
👍 👍

PR #134, opened 2026-05-11
fix: Include broker connection stats in adapter report payload

adamlogic:
## Summary

- Snapshots `connected_clients` and `maxclients` from the broker Redis via `INFO clients` at the top of every `CeleryMetricsCollector.collect()` cycle and ships them under a new `broker` block on the celery adapter's entry in the outgoing report.
- Logs a warning whenever the broker reports fewer than 10 free connection slots, so customers can spot connection-cap pressure before `inspect.active()` starts failing.
- Adds a `report_metadata` property to the `Collector` protocol (default `{}`) and merges it into `Adapter.as_tuple` so any collector can contribute adapter-scoped fields without further wiring.
- Purpose: surface when a broker is approaching its connection cap. On TLS Redis that ceiling otherwise manifests as an opaque `SSLEOFError: UNEXPECTED_EOF_WHILE_READING` during a fresh connection's handshake (Redis closes the socket before any application-layer error can be sent), instead of the clean `max number of clients reached` you'd see over plain TCP.

JDO-1363

## Cost

`INFO clients` is one round trip on the already-open broker connection. Benchmarked locally at ~70-100µs server-side; expect ~1-2ms total in-region. The collector already calls `INFO` (all sections) at startup for the version check, so this isn't a new permission requirement on managed Redis providers.

## Payload shape

Old:
```json
"judoscale-celery": {"runtime_version": "5.6.3", "adapter_version": "1.13.2"}
```

New (when stats are available):
```json
"judoscale-celery": {
  "runtime_version": "5.6.3",
  "adapter_version": "1.13.2",
  "broker": {"connected_clients": 31, "maxclients": 40}
}
```

The `broker` block is omitted entirely if the `INFO` call fails or returns missing keys, so the report never carries stale or partial numbers. Additive change to the payload — older agents simply don't send it.

## Warning behavior

When `maxclients - connected_clients < 10`:

```
WARNING - [judoscale] Broker is near its connection limit: 31/40 connections in use (9 remaining). New connections may be rejected, which on TLS Redis can surface as SSL handshake errors.
```

One warning per report cycle (every 10s by default). Stays silent when headroom is healthy.

## Test plan

- [x] `poetry run pytest` — 151 passed (6 new tests covering the broker-stats lifecycle, warning threshold, and adapter merge behavior).
- [x] Manual end-to-end against a local TLS Redis in Docker, sized so the first reporter cycle succeeds and the second one trips the cap: verified the `broker` block makes it onto the payload on both cycles, the warning fires whenever `< 10` slots remain, and `inspect.active()`'s `SSLEOFError` on cycle 2 is absorbed by the JDO-1362 collector-survival wrapper without losing the broker metadata.
- [ ] Backend coordination: the receiving adapter API needs to know how to parse and store the new `broker` field. Tracked separately.

---

> [!NOTE]
> **Medium Risk**
> Adds new top-level `metadata` to the report payload and introduces Redis `INFO clients` calls/logging in the Celery collector; could affect downstream API parsing and increase broker load/log volume if misconfigured.
> 
> **Overview**
> Reports now include a new top-level `metadata` block, built by merging a new `report_metadata` property exposed by all metrics collectors.
> 
> `CeleryMetricsCollector` now snapshots Redis broker connection stats (`connected_clients`, `maxclients`) via `INFO clients` each `collect()` cycle, publishes them under `metadata['celery-broker']`, and logs when remaining connection headroom drops below a threshold.
> 
> Tests were updated/added to cover section-aware Redis `INFO` mocking, metadata inclusion/omission behavior, and the near-connection-limit log trigger.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8e4b9717c4a92ef7a4d7dbaef0be465916bbde99. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/judoscale/celery/collector.py
+++ b/judoscale/celery/collector.py
@@ -21,6 +21,12 @@ DEFAULTS = {
     "TRACK_BUSY_JOBS": False,
 }
 
+# Warn when the broker has fewer than this many free connection slots.
+# Crossing into single digits is a strong predictor of pidbox failures
+# (e.g. the SSLEOFError seen on TLS Redis when new connections are
+# rejected mid-handshake under cap exhaustion).
+BROKER_CONNECTIONS_WARN_THRESHOLD = 10
+
 
 class TaskSentHandler(Thread):
     def __init__(
@@ -115,6 +121,17 @@ class CeleryMetricsCollector(JobMetricsCollector):
                 except (TypeError, ValueError):
                     continue
         self._broker_stats = stats
+
+        if "connected_clients" in stats and "maxclients" in stats:
+            remaining = stats["maxclients"] - stats["connected_clients"]
+            if remaining < BROKER_CONNECTIONS_WARN_THRESHOLD:
+                logger.warning(
+                    f"Broker is near its connection limit: "
+                    f"{stats['connected_clients']}/{stats['maxclients']} "
+                    f"connections in use ({remaining} remaining). "
+                    f"New connections may be rejected, which on TLS Redis "
+                    f"can surface as SSL handshake errors."
+                )
 
     @property
     def _queues(self) -> List[str]:
--- a/judoscale/celery/collector.py
+++ b/judoscale/celery/collector.py
@@ -21,11 +21,11 @@ DEFAULTS = {
     "TRACK_BUSY_JOBS": False,
 }
 
-# Warn when the broker has fewer than this many free connection slots.
-# Crossing into single digits is a strong predictor of pidbox failures
+# Log when the broker has fewer than this many free connection slots.
+# Running low on headroom is a strong predictor of pidbox failures
 # (e.g. the SSLEOFError seen on TLS Redis when new connections are
 # rejected mid-handshake under cap exhaustion).
-BROKER_CONNECTIONS_WARN_THRESHOLD = 10
+BROKER_CONNECTIONS_INFO_THRESHOLD = 20
 
 
 class TaskSentHandler(Thread):
@@ -128,8 +128,8 @@ class CeleryMetricsCollector(JobMetricsCollector):
 
         if "connected_clients" in stats and "maxclients" in stats:
             remaining = stats["maxclients"] - stats["connected_clients"]
-            if remaining < BROKER_CONNECTIONS_WARN_THRESHOLD:
-                logger.warning(
+            if remaining < BROKER_CONNECTIONS_INFO_THRESHOLD:
+                logger.info(
                     f"Broker is near its connection limit: "
                     f"{stats['connected_clients']}/{stats['maxclients']} "
                     f"connections in use ({remaining} remaining). "
--- a/judoscale/celery/collector.py
+++ b/judoscale/celery/collector.py
@@ -112,9 +112,13 @@ class CeleryMetricsCollector(JobMetricsCollector):
             self._broker_stats = {}
             return
 
+        if not isinstance(info, dict):
+            self._broker_stats = {}
+            return
+
         stats: Dict[str, int] = {}
         for key in ("connected_clients", "maxclients"):
-            value = info.get(key) if isinstance(info, dict) else None
+            value = info.get(key)
             if value is not None:
                 try:
                     stats[key] = int(value)
--- a/judoscale/celery/collector.py
+++ b/judoscale/celery/collector.py
@@ -3,7 +3,7 @@ import time
 from datetime import datetime
 from collections import defaultdict
 from threading import Thread
-from typing import List, Optional, Set, Tuple
+from typing import Dict, List, Optional, Set, Tuple
 
 from celery import Celery
 from kombu import Connection
@@ -71,6 +71,7 @@ class CeleryMetricsCollector(JobMetricsCollector):
 
         self._celery_queues: Set[str] = set()
         self._queues_scanned: bool = False
+        self._broker_stats: Dict[str, int] = {}
         self.task_sent_handler = TaskSentHandler(self, connection)
         logger.debug(f"Redis is at {self.redis.connection_pool}")
         self.task_sent_handler.start()
@@ -83,6 +84,37 @@ class CeleryMetricsCollector(JobMetricsCollector):
     @property
     def adapter_config(self):
         return self.config["CELERY"]
+
+    @property
+    def report_metadata(self) -> dict:
+        # Only emit the broker block when we have fresh stats from the most
+        # recent collect() cycle; suppress it entirely on cycles where the
+        # INFO call failed so the report doesn't carry stale numbers.
+        if not self._broker_stats:
+            return {}
+        return {"broker": dict(self._broker_stats)}
+
+    def _refresh_broker_stats(self) -> None:
+        """
+        Snapshot `connected_clients` and `maxclients` from the broker Redis
+        via `INFO clients` for inclusion in the next report.
+        """
+        try:
+            info = self.redis.info("clients")
+        except Exception as e:
+            logger.debug(f"Unable to refresh broker stats: {e}")
+            self._broker_stats = {}
+            return
+
+        stats: Dict[str, int] = {}
+        for key in ("connected_clients", "maxclients"):
+            value = info.get(key) if isinstance(info, dict) else None
+            if value is not None:
+                try:
+                    stats[key] = int(value)
+                except (TypeError, ValueError):
+                    continue
+        self._broker_stats = stats
 
     @property
     def _queues(self) -> List[str]:
@@ -139,6 +171,8 @@ class CeleryMetricsCollector(JobMetricsCollector):
         metrics = []
         if not self.should_collect:
             return metrics
+
+        self._refresh_broker_stats()
 
         if not self._queues_scanned:
             self._scan_queues()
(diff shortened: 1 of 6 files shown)


judoscale/core/adapter.py
```diff
@@ -25,4 +25,10 @@ class Adapter:
 
     @property
     def as_tuple(self):
-        return (self.identifier, asdict(self.adapter_info))
+        info = asdict(self.adapter_info)
+        if self.metrics_collector is not None:
+            # Let the collector contribute adapter-scoped fields (broker
+            # stats, etc.) computed during its most recent collect() cycle.
+            extra = getattr(self.metrics_collector, "report_metadata", None) or {}
```
carlosantoniodasilva: We introduced the concept of "metadata" added to each report, which is stored separately: I had applied it to a previous PR for troubleshooting but it never got merged, but the infra should still be in place.

https://github.com/judoscale/judoscale-python/pull/114/changes#diff-dc3605e76b7fa7403b885af969869ff577f3020b0bea517455a47caeb6bf745aR131-R135

Applying this to the adapter info itself means it gets mangled with adapters and will cause additional updates on the backend as the adapters get propagated, so it might be best if we can keep it out of that. (adapters is more "stable" and only changes with version changes, metadata is more volatile since it is likely going to be different with each report)

judoscale/celery/collector.py
```diff
@@ -21,6 +21,12 @@
     "TRACK_BUSY_JOBS": False,
 }
 
+# Warn when the broker has fewer than this many free connection slots.
+# Crossing into single digits is a strong predictor of pidbox failures
+# (e.g. the SSLEOFError seen on TLS Redis when new connections are
+# rejected mid-handshake under cap exhaustion).
+BROKER_CONNECTIONS_WARN_THRESHOLD = 10
```
carlosantoniodasilva: I wonder if this should be a % of connections left, e.g. if you have 40 connections than maybe 4, if you have 100 than it's 10. (assuming we keep a 10% threshold)

judoscale/celery/collector.py
```diff
@@ -84,6 +91,48 @@ def is_supported_redis_version(self):
...
         return self.config["CELERY"]
 
+    @property
+    def report_metadata(self) -> dict:
+        # Only emit the broker block when we have fresh stats from the most
+        # recent collect() cycle; suppress it entirely on cycles where the
+        # INFO call failed so the report doesn't carry stale numbers.
+        if not self._broker_stats:
+            return {}
+        return {"broker": dict(self._broker_stats)}
```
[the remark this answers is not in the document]
carlosantoniodasilva: This seems fine.

judoscale/celery/collector.py
```diff
@@ -84,6 +91,48 @@ def is_supported_redis_version(self):
...
+        try:
+            info = self.redis.info("clients")
+        except Exception as e:
+            logger.debug(f"Unable to refresh broker stats: {e}")
+            self._broker_stats = {}
+            return
+
+        stats: Dict[str, int] = {}
+        for key in ("connected_clients", "maxclients"):
+            value = info.get(key) if isinstance(info, dict) else None
```
carlosantoniodasilva: It feels like this `if isinstance(info, dict)` could be pushed up, since it is a guard that has nothing to do with the value check? It's fine as-is though, just a thought.

tests/test_collectors.py
```diff
@@ -358,6 +369,86 @@ def test_collect_with_busy_jobs(self, worker_1, celery, monkeypatch):
...
+        assert collector.report_metadata == {}
+
+    def test_report_metadata_empty_when_info_missing_keys(self, worker_1, celery):
+        redis_client = celery.connection_for_read().channel().client
+        redis_client.scan_iter.return_value = []
+
+        collector = CeleryMetricsCollector(worker_1, celery)
+        # Some managed Redis providers may strip keys from INFO clients;
+        # in that case we report no broker block rather than half a one.
+        redis_client.info.side_effect = lambda section=None: {}
```
carlosantoniodasilva, line 412: I guess what bugbot was saying is that if the Redis returns something like `{"connected_clients": X}` without for example `"max_clients"`, we'd pass only that to the report metadata... which I guess is fine.

tests/test_collectors.py
```diff
@@ -358,6 +369,86 @@ def test_collect_with_busy_jobs(self, worker_1, celery, monkeypatch):
...
+        collector = CeleryMetricsCollector(worker_1, celery)
+        assert collector.report_metadata == {}
+
+    def test_report_metadata_populated_after_collect(self, worker_1, celery):
+        celery.connection_for_read().channel().client.scan_iter.return_value = []
+        collector = CeleryMetricsCollector(worker_1, celery)
+        collector.collect()
+        assert collector.report_metadata == {
+            "broker": {"connected_clients": 3, "maxclients": 40}
+        }
```
carlosantoniodasilva: Feels like these could be merged maybe.

```suggestion
    def test_report_metadata_populates_after_collect(self, worker_1, celery):
        celery.connection_for_read().channel().client.scan_iter.return_value = []
        collector = CeleryMetricsCollector(worker_1, celery)
        assert collector.report_metadata == {}

        collector.collect()
        assert collector.report_metadata == {
            "broker": {"connected_clients": 3, "maxclients": 40}
        }
```

tests/test_collectors.py
```diff
@@ -358,6 +369,86 @@ def test_collect_with_busy_jobs(self, worker_1, celery, monkeypatch):
...
+        collector = CeleryMetricsCollector(worker_1, celery)
+        collector.collect()
+
+        assert any(
+            "Broker is near its connection limit" in record.message
+            and "31/40" in record.message
+            for record in caplog.records
+        )
+
+    def test_does_not_warn_when_broker_has_headroom(
```
carlosantoniodasilva: We could also maybe test this one as part of the base "populate" test, since it's all the same setup, just a matter of checking no log output. But it's fine on its own too to maybe not make that one too noisy.

carlosantoniodasilva (commented):
Looks good, I had a couple thoughts below, mainly about whether to expand on the adapters key vs exposing this as separate metadata.

Do you plan on shipping and keeping this, or is it more of a potential temp troubleshooting change?

judoscale/celery/collector.py
adamlogic: I went with a static number because our connection-usage is not dependent on the total max clients. Our "busy job tracking" can use up to 18 clients in my testing. (Details in [Linear comment](https://linear.app/judo/issue/JDO-1363/investigate-celery-track-busy-jobs-ssl-failure-on-heroku-tls-redis#comment-111db85a))

But as for why I chose "10" as the number... I'm not really sure. Since I'm seeing 18 clients in my testing, maybe 20 makes more sense here.

tests/test_collectors.py
adamlogic: Yeah, I think so.

judoscale/core/adapter.py
adamlogic: Good call, I forgot we supported that.

adamlogic (commented):
> Do you plan on shipping and keeping this, or is it more of a potential temp troubleshooting change?

Was thinking it'd be helpful to have permanently.

judoscale/celery/collector.py
adamlogic: I think I want to stick with 10... 20 feels too aggressive since Heroku's cheapest Redis plans have very restrictive limits. We'll only _potentially_ use more than 10 connections if the client is configured to track busy jobs, so I feel like 10 hits the spot between the two. 🤷‍♂️ Not a strong opinion on it.

adamlogic (commented):
Thanks for the suggestions! Ready for another look.

judoscale/celery/collector.py
carlosantoniodasilva: Sounds good, it's a starting point we can tweak going forward.

It's really odd how many clients/connections tracking busy jobs can take tbh.

judoscale/celery/collector.py
```diff
@@ -84,6 +91,52 @@ def is_supported_redis_version(self):
...
         return self.config["CELERY"]
 
+    @property
+    def report_metadata(self) -> dict:
+        # Only emit the broker block when we have fresh stats from the most
+        # recent collect() cycle; suppress it entirely on cycles where the
+        # INFO call failed so the report doesn't carry stale numbers.
+        if not self._broker_stats:
+            return {}
+        return {"broker": dict(self._broker_stats)}
```
carlosantoniodasilva: Since this goes in a "generic metadata" field, maybe we prefix with `celery`, e.g.:

```suggestion
        return {"celery-broker": dict(self._broker_stats)}
```

or similar... just in case we add more stuff to metadata from other packages in the future. (for example dramatiq also has the concept of a "broker")

carlosantoniodasilva (approved):
Just another small suggestion, but otherwise LGTM! 👍

judoscale/celery/collector.py
adamlogic: > 20 feels too aggressive

I'm changing my mind on this. I think 20 is a better starting point based on what I'm seeing with a current customer support thread. I'm going to change the logging from WARN to INFO though so it's easier to silence (it'll going to log every 10 seconds).

adamlogic (commented):
Thanks!

PR #138, opened 2026-06-12
refactor: Model the hosting platform with per-platform classes

adamlogic:
## Summary
Ports [judoscale/judoscale-ruby#275](https://github.com/judoscale/judoscale-ruby/pull/275).

- Replace the string-based `RuntimeContainer` with a `Platform` hierarchy detected from the environment, so platform-specific behavior lives on the classes that actually have those concepts instead of being re-derived from container id shape.
- Heroku and Scalingo own ordinal redundancy detection; opaque-id platforms (Render, ECS, Fly, Railway, etc.) default to non-redundant. Render also strips the service-id prefix and supplies a legacy `default_api_base_url` when `JUDOSCALE_URL` is unset.
- Call sites switch from `RUNTIME_CONTAINER` to `config["PLATFORM"]`; the reported `container` JSON key is unchanged.

---

> [!NOTE]
> **Medium Risk**
> Touches metric reporting enablement (release/redundant instances) and legacy Render API URL derivation; behavior is intended to match prior logic with clearer platform boundaries and expanded ordinal tests.
> 
> **Overview**
> Replaces the string-based **`RuntimeContainer`** with a **`Platform`** hierarchy detected via **`Platform.detect(env)`**, so Heroku/Scalingo ordinal redundancy, Heroku release dynos, and Render legacy API URL behavior live on platform subclasses instead of regex on container id shape.
> 
> **`Config`** now stores **`PLATFORM`** (not **`RUNTIME_CONTAINER`**), delegates env parsing to **`Platform.detect`**, and fills **`API_BASE_URL`** from **`platform.default_api_base_url`** when **`JUDOSCALE_URL`** is unset. **Reporter** and **job metrics collectors** read **`config["PLATFORM"]`** for release skip, redundant-instance skip, and report **`container`** (unchanged JSON key). Ordinal matching no longer caps at three digits (e.g. **`web.1000`** is redundant).
> 
> Tests and sample apps updated; **`TestPlatform`** replaces **`TestRuntimeContainer`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5fe11aca6a23d5c87d91685245427d0f02477681. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/judoscale/core/platform.py
+++ b/judoscale/core/platform.py
@@ -1,0 +1,128 @@
+import os
+import re
+from typing import Mapping
+
+
+class Platform:
+    """The hosting platform we detected from the environment.
+
+    The container/instance id is just one property of the platform — behavior
+    that only applies to certain platforms (whether an instance is a redundant
+    member of a formation, or a release/build instance) lives on the platform
+    subclasses that actually have those concepts, instead of being re-derived
+    from the shape of the container string.
+    """
+
+    def __init__(self, container):
+        self.container = str(container)
+
+    @property
+    def is_redundant_instance(self) -> bool:
+        """Most platforms expose opaque, non-ordinal instance ids (Render, ECS,
+        Fly, Railway, custom), so by default no instance is redundant. Platforms
+        that have a formation ordinal override this."""
+        return False
+
+    @property
+    def is_release_instance(self) -> bool:
+        """Only Heroku has a release/build phase, so by default no instance is a
+        release instance."""
+        return False
+
+    @property
+    def default_api_base_url(self):
+        """Platforms may contribute a default API base url when one isn't configured."""
+        return None
+
+    @classmethod
+    def detect(cls, env: Mapping = os.environ) -> "Platform":
+        """Detect the current platform from the environment. Order matters: an
+        explicit JUDOSCALE_CONTAINER always wins, and Unknown is the fallback."""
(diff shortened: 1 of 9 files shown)


judoscale/core/platform.py
```diff
@@ -0,0 +1,128 @@
...
+
+    @property
+    def is_redundant_instance(self) -> bool:
+        match = self._ORDINAL.fullmatch(self.container)
+        return int(match.group(1)) > 1 if match else False
+
+    @property
+    def is_release_instance(self) -> bool:
+        # Heroku release-phase dynos are named "release.1234".
+        return self.container.lower().startswith("release")
```
carlosantoniodasilva, line 76: I think the Python package seems to be the only one which checks for this too, and we don't have the one-off container check. (if we're looking for more consistency across them all)

carlosantoniodasilva (approved):
This all looks great! It's a bit more boilerplate code, but it feels much better organized and easier to follow.

Just one note about potential consistency across the board (release check only here, not having the one-off check.)

judoscale/core/platform.py
adamlogic: 👍 will follow up

PR #140, opened 2026-06-12
chore: Upgrade Black for Python 3.14

adamlogic:
## Summary

- Upgrade Black to a Python 3.14-compatible version.
- Refresh the Poetry lockfile for Black's updated dependency set.
- Apply the formatting changes produced by the upgraded formatter.

## Design decisions

This PR is based on `cursor/one-off-container-check` so the formatter compatibility fix stays separate from the ephemeral-container behavior changes while still applying cleanly on top of that branch.

The upgrade resolves the Python 3.14 formatter failure caused by older Black relying on legacy `ast` aliases that no longer exist.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/judoscale/core/__init__.py
+++ b/judoscale/core/__init__.py
@@ -5,6 +5,7 @@ from judoscale.core.reporter import reporter
 
 previous_handler = signal.getsignal(signal.SIGTERM)
 
+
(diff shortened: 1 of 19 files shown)


carlosantoniodasilva (approved):
👍

